Issue #24, opened 2022-03-21
Fix failing tests


gcroci2:
Lots of tests are failing. For details see [here](https://github.com/DeepRank/deeprank-gnn-2/runs/5624296084?check_suite_focus=true)

PR #29, opened 2022-03-22
Add type hinting to models module

svenvanderburg:
Adds type hinting to `models` module. 

- [x] models/amino_acid
- [ ] models/conservation , see #32 
- [x] models/error (only contains one error message)
- [x] models/graph
- [x] models/pair 
- [x] models/polarity (just an enum)
- [x] models/query
- [x] models/structure
- [x] models/forcefield/

I tried to run the tests locally. I have 9 failing tests, but I have the same results on the `main` branch:
```
FAILED tests/test_nn.py::TestNeuralNet::test_fout - IndexError: list index out of range
FAILED tests/test_nn.py::TestNeuralNet::test_ginet - IndexError: list index out of range
FAILED tests/test_nn.py::TestNeuralNet::test_ginet_class - IndexError: list index out of range
FAILED tests/test_nn.py::TestNeuralNet::test_sgat - IndexError: list index out of range
FAILED tests/test_nn.py::TestNeuralNet::test_simple - IndexError: list index out of range
FAILED tests/test_nn.py::TestNeuralNet::test_simple_ginet - IndexError: list index out of range
FAILED tests/test_preprocess.py::test_preprocess - AssertionError: 100 != 94
FAILED tests/domain/test_amino_acid.py::test_all_different_onehot - TypeError: unhashable type: 'numpy.ndarray'
FAILED tests/tools/test_pdb.py::test_get_structure_from_nmr_with_dna - FileNotFoundError: File not found: tests/data/pdb/1A6B/1A6B.pdb
============================================================================================= 9 failed, 35 passed, 4 xfailed, 1873 warnings in 181.21s (0:03:01)
```

So I assume I didn't break anything, of course to be 100% sure we need these tests to be fixed (probably with #24 ).


deeprank_gnn/models/query.py
```diff
@@ -1,25 +1,27 @@
...
-
+from deeprank_gnn.domain.graph import EDGETYPE_INTERNAL, EDGETYPE_INTERFACE
+from deeprank_gnn.models.error import UnknownAtomError
+from deeprank_gnn.models.forcefield.vanderwaals import VanderwaalsParam
+from deeprank_gnn.models.graph import Graph
+from deeprank_gnn.models.structure import Residue, Atom
+from deeprank_gnn.tools import BioWrappers, BSA
+from deeprank_gnn.tools.pdb import (get_residue_contact_pairs, get_residue_distance, get_surrounding_residues,
+                                    get_structure)
+from deeprank_gnn.tools.pssm import parse_pssm
```
svenvanderburg, line 24: I reordered this according to best Python practices.

deeprank_gnn/models/forcefield/top.py
```diff
@@ -1,7 +1,7 @@
 
 
 class TopRowObject:
-    def __init__(self, residue_name, atom_name, kwargs):
+    def __init__(self, residue_name: str, atom_name: str, kwargs: dict):
```
cbaakman, line 4: shouldn't kwargs become Dict[str, str] ? Or is this the default?

deeprank_gnn/models/graph.py
```diff
@@ -1,11 +1,12 @@
+from typing import Dict
 from uuid import uuid4
 
 from networkx import Graph as NetworkxGraph
 
 class Graph(NetworkxGraph):
     "this is a graph just like in networkx, but with an id and associated target values"
 
-    def __init__(self, id_=None, targets=None):
+    def __init__(self, id_: str = None, targets: Dict[str, float] = None):
```
cbaakman, line 9: Shouldn't the optional arguments be set to type Optional?

deeprank_gnn/models/structure.py
```diff
@@ -133,23 +138,23 @@ def get_pssm(self):
...
     def chain(self):
         return self._chain
 
     @property
-    def amino_acid(self):
+    def amino_acid(self) -> AminoAcid:
         return self._amino_acid
 
     @property
     def atoms(self):
```
cbaakman, line 153: What about the return type here?

cbaakman (approved):

deeprank_gnn/models/forcefield/top.py
svenvanderburg: Good point. I changed it to `Dict[str, any]`

deeprank_gnn/models/graph.py
svenvanderburg: Yes! I changed this everywhere in the models folder.

deeprank_gnn/models/structure.py
svenvanderburg: That's a good one 😉  It should be `List[Atom]`, but we haven't defined `Atom` here yet (we do that below in the `structure` module). The proper way to do this would be to work with interfaces (i.e. an `AtomInterface` that we define in a separate module), but I didn't go for it because it will increase the complexity of the code and I am not sure if we will have much benefits from it. Another way would be to reorder the class definitions in `models/structure.py` so that `Atom` is defined before `Residue`. But then we have the same problem for `Residue.chain`.

PR #34, opened 2022-03-23
Fix metadata in setup.py

gcroci2:
Metadata has been updated with the current repository and authors' information.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,20 +15,20 @@ with open('README.md') as readme_file:
     readme = readme_file.read()
 
 setup(
-    name='DeepRank-GNN',
+    name='DeepRank-GNN-2',
     version=version['__version__'],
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -8,27 +8,27 @@
...
 
 with open('README.md') as readme_file:
     readme = readme_file.read()
 
 setup(
-    name='DeepRank-GNN',
+    name='DeepRank-GNN-2',
     version=version['__version__'],
-    description='Graph Neural network Scoring of protein-protein conformations',
+    description='TBD',
```
svenvanderburg, line 20: I would just leave the current description, probably seems suitable for deeprank-gnn-2 as well.

svenvanderburg (approved):
Just a few comments, but feel free to merge after addressing them! :)

PR #46, opened 2022-03-29
Use coveralls github action

svenvanderburg:


svenvanderburg:
The github action does not work with pytest, original solution actually works now, see https://coveralls.io/github/DeepRank/deeprank-gnn-2

Issue #56, opened 2022-04-13
Replace multiprocessing.Queue with multiprocessing.pool.map in PreProcessor


svenvanderburg:
The `PreProcessor` class currently uses a quite low-level combination of `multiprocessing.Process` and `multiprocessing.Queue` to parallelise graph construction from .pdb files based on queries.

See https://github.com/DeepRank/deeprank-gnn-2/blob/c0bff28d73bf67a14fd4f55d9adac5884a2dde1a/deeprank_gnn/preprocess.py#L73

From what I understand, we don't really need queues. This is more a map-type problem. Because we always just have a list of queries we want to send at once to the preprocessor, and then just want to collect the results. (for example we are not polling from another process and then adding to the queue).

I suggest to refactor `Preprocessor` so that it makes use of `multiprocessing.pool.map` instead. Benefits:
* More high-level, code will be simplified a lot
* We don't have to worry about the different processes itself, that is all handled for us.
* This will also solve #55, and guarantees to solve the problem that arose in #47 and was presumably fixed with using `Queue.qsize()`. (i.e. checking whether the process is done is handled by `pool.map` itself)

svenvanderburg:
@cbaakman can you comment on whether you think this is a good idea? (or maybe we can discuss)

cbaakman:
It's a good idea to use a pool. But I don't have experience with that. It doesn't seem to have priority to me, if the current approach works as well.

PR #57, opened 2022-04-19
Use workaround for qsize not implemented on Mac OSX

### LINKED ISSUE #55: Queue.qsize() NotImplementedError macOS because of broken sem_getvalue()
When running `test_process()` it fails on MacOS with:
```
tests/test_preprocess.py:35:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
deeprank_gnn/preprocess.py:109: in wait
    while self._input_queue.qsize() > 0:  # qsize seems to be more reliable than the empty method
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <multiprocessing.queues.Queue object at 0x16db95880>

    def qsize(self):
        # Raises NotImplementedError on Mac OSX because of broken sem_getvalue()
>       return self._maxsize - self._sem._semlock._get_value()
E       NotImplementedError
```

svenvanderburg:
Fixes #55 

This at least does an attempt to have something working on OSX, although we know that when using `.empty()` we already stop while some jobs are still running. This will be fixed with #56.

### DIFF
--- a/deeprank_gnn/preprocess.py
+++ b/deeprank_gnn/preprocess.py
@@ -101,12 +101,24 @@ class PreProcessor:
             if not process.is_alive():
                 raise RuntimeError(f"worker process {process.name} did not start")
 
+    def _queue_empty(self):
+        """
+        Returns whether queue is empty.
+
+        Note: This implementation is not reliable, especially not on Mac OSX .qsize() seems to be more reliable than
+        .empty(), however qsize relies on sem_getvalue() which is not implemented on Mac OSX.
+        """
+        try:
+            return self._input_queue.qsize() == 0
+        except NotImplementedError:
+            return self._input_queue.empty()
+
     def wait(self):
         "wait for all graphs to be built"
 
         try:
             _log.info("waiting for the queue to be empty..")
-            while self._input_queue.qsize() > 0:  # qsize seems to be more reliable than the empty method
+            while not self._queue_empty():
                 sleep(1.0)
         finally:
             self.shutdown()
--- a/deeprank_gnn/preprocess.py
+++ b/deeprank_gnn/preprocess.py
@@ -105,7 +105,7 @@ class PreProcessor:
         """
         Returns whether queue is empty.
 
-        Note: This implementation is not reliable, especially not on Mac OSX .qsize() seems to be more reliable than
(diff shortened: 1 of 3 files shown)


deeprank_gnn/preprocess.py
```diff
@@ -101,12 +101,24 @@ def start(self):
...
+        """
+        Returns whether queue is empty.
+
+        Note: This implementation is not reliable, especially not on Mac OSX. .qsize() seems to be more reliable than
+        .empty(), however qsize relies on sem_getvalue() which is not implemented on Mac OSX.
+        """
+        try:
+            return self._input_queue.qsize() == 0
+        except NotImplementedError:
+            return self._input_queue.empty()
```
cbaakman, line 114: Are you sure that the preprocessing still works on a MAC, if "empty" is used?

svenvanderburg: Well, it works, as in: it runs. But it is not reliable, as in: we don't wait long enough for all processes to finish. (I hope that is clear from the documentation that I added, if not, let me know how I can improve it). 

But I think it would be good to **properly** fix this with #56. The question is mostly: how much do you care about Mac OSX users haha.

cbaakman: I suppose we don't have other options right now. Let's just merge this as soon as the checks complete.

PR #59, opened 2022-04-21
Fix prospector errors in domain folder

gcroci2:


deeprank_gnn/domain/amino_acid.py
```diff
@@ -1,31 +1,189 @@
-from deeprank_gnn.models.amino_acid import AminoAcid
-from deeprank_gnn.models.polarity import Polarity
+from ..models.amino_acid import AminoAcid
```
gcroci2, line 1: Relative imports only work within packages, so I added __init__.py file in domain folder so that then domain folder becomes a package.

cbaakman (approved):

PR #62, opened 2022-04-25
fix prospector errors in models/ main dir

gcroci2:


deeprank_gnn/models/conservation.py
```diff
@@ -22,8 +20,11 @@ def get_conservation(self, amino_acid):
 class ConservationTable:
     "holds data for one position-specific scoring table"
 
-    def __init__(self, rows={}):
-        self._rows = rows
+    def __init__(self, rows=None):
```
gcroci2, line 23: It's dangerous to use an empty dict as the default value. If you modify a default argument, it will persist until the next call, so your "empty" dict will start to contain values on calls other than the first one.

deeprank_gnn/models/contact.py
```diff
@@ -25,15 +22,19 @@ def vanderwaals_potential(self) -> float:
...
+        distance: float,
+        electrostatic_potential: float,
+        vanderwaals_potential: float,
+    ):
 
         self._residue1 = residue1
         self._residue2 = residue2
 
-        Pair.__init__(self, residue1, residue2)
+        # Pair.__init__(self, residue1, residue2)
```
gcroci2, line 37: Is this line supposed to do anything?

deeprank_gnn/models/contact.py
```diff
@@ -53,25 +54,29 @@ def vanderwaals_potential(self) -> float:
 
     @property
     def residue1(self) -> Residue:
-        return self.item1
+        return self.residue1
```
gcroci2, line 57: I think this is the right attribute (item1 does not exist)

deeprank_gnn/models/contact.py
```diff
@@ -53,25 +54,29 @@ def vanderwaals_potential(self) -> float:
...
 
     @property
     def residue2(self) -> Residue:
-        return self.item2
+        return self.residue2
```
gcroci2, line 61: I think this is the right attribute (item2 does not exist)

deeprank_gnn/models/graph.py
```diff
@@ -109,7 +122,7 @@ def add_edge(self, edge: Edge):
         self._edges[edge.id] = edge
 
     def get_edge(self, id_: Contact) -> Edge:
-        return self._edges[edge.id]
+        return self._edges[id_]
```
gcroci2, line 125: To be sure, confirm this.

deeprank_gnn/models/grid.py
```diff
@@ -118,54 +127,69 @@ def features(self) -> Dict[str, numpy.array]:
...
+    # ) -> numpy.ndarray:
 
-        order = 4
+    #     order = 4
 
-        fx, fy, fz = position
-        bsp_data = (bspline((self.xgrid - fx) / self.resolution, order) *
-                    bspline((self.ygrid - fy) / self.resolution, order) *
-                    bspline((self.zgrid - fz) / self.resolution, order))
+    #     fx, fy, fz = position
```
gcroci2, line 179: I commented the BSP line everywhere since we did not know where bspline came from.

deeprank_gnn/models/query.py
```diff
@@ -453,76 +649,111 @@ def _set_amino_acid(graph: Graph, variant_residue: Residue,
...
 
             node.features[FEATURENAME_INFORMATIONCONTENT] = pssm_row.information_content
             node.features[FEATURENAME_PSSM] = pssm_value
 
     @staticmethod
     def _set_sasa(graph: Graph, pdb_path: str):
 
-        structure = freesasa.Structure(pdb_path)
-        result = freesasa.calc(structure)
+        structure = freesasa.Structure(pdb_path) # pylint: disable=c-extension-no-member
```
gcroci2, line 713: Pylint imports modules to effectively identify valid methods and attributes. It was decided that importing c extensions that are not part of the python stdlib is a security risk and could introduce malicious code. So pylint gives me errors when we use freesasa (that is a C-library). Can this be a concern at some point?

deeprank_gnn/models/error.py
```diff
@@ -1,5 +1,3 @@
-
-
-class UnknownAtomError(ValueError):
-    "should be raised when an unknown atom type is encountered"
-    pass
+class UnknownAtomError(ValueError):
```
gcroci2, line 1: we should either implement something here or remove this class everywhere.

deeprank_gnn/models/contact.py
cbaakman: it initializes the parent class.

deeprank_gnn/models/contact.py
cbaakman: I suppose _residue1 will work, but not residue1, since that would be a circular call.

deeprank_gnn/models/contact.py
cbaakman: I suppose _residue2 will work, but not residue2, since that would be a circular call.

deeprank_gnn/models/graph.py
cbaakman: This is right.

deeprank_gnn/models/graph.py
```diff
@@ -1,16 +1,21 @@
...
 
 import numpy
 import h5py
 
-from deeprank_gnn.models.structure import Atom, Residue
-from deeprank_gnn.models.contact import Contact
-from deeprank_gnn.models.grid import MapMethod, Grid, GridSettings
-from deeprank_gnn.domain.storage import *
-
+from .structure import Atom, Residue
```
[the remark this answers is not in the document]
cbaakman: You can't just remove the imports, they are used within the script.
Have you ran the tests/model/test_graph.py ?

deeprank_gnn/models/grid.py
cbaakman: I think I forgot the

**from scipy.signal import bspline**

deeprank_gnn/models/query.py
cbaakman: Deeprank runs on a dedicated server, where it will do calculations. It will not need root rights to do this.
SASA computations are necessary for deeprank to function. If you find an alternative, it would be acceptable to drop SASA.

deeprank_gnn/models/error.py
cbaakman: In deeprank, we take a specific action when this specific error is thrown. All other errors are considered unexpected.
We need to know when an unknown atom is encountered, so that we can skip it.
It doesn't matter whether this class implements something, as long as it can be identified as an unknown atom error.

deeprank_gnn/models/contact.py
gcroci2: Oh, I see. But the parent class of ResidueContact is Contact, not Pair, right? Pair is parent class for Contact instead. I think we should use `super()` (see [here](https://docs.python.org/3/library/functions.html#super)) both in Contact and ResidueContact `__init__` methods

deeprank_gnn/models/contact.py
gcroci2: In this case either we change `residue1` method name or the attribute name. Anyway, `item1` is not an attribute at all of this class, right? Maybe we can use `item1` and `item2` as attributes names (changing them also in the init method), and leave `residue1` and `residue2` methods?

deeprank_gnn/models/graph.py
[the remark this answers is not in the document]
gcroci2: Indeed I removed only the unused imports. And instead of importing all functions/classes from modules (with `import *`), I copied explicitly functions/classes and then I removed only the ones not used (see the lines of code below)

deeprank_gnn/models/grid.py
gcroci2: Ok I'll insert it

deeprank_gnn/models/query.py
gcroci2: Ok so for now it's perfectly fine

deeprank_gnn/models/error.py
gcroci2: What about printing the ValueError?

deeprank_gnn/models/contact.py
cbaakman: I made Contact inherit from Pair, since it's a pair of residues and their order should be irrelevant. However, Pair has the fields item1 & item2, not residue1 & residue2. An alternative would be not to inherit from Pair and copy some class methods from Pair instead.

deeprank_gnn/models/conservation.py
cbaakman: You're right!

deeprank_gnn/models/graph.py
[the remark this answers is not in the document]
cbaakman: Sorry! I didn't seed the part below.

deeprank_gnn/models/error.py
cbaakman: You mean that we should let python check the error message in the catch block?

gcroci2: Sorry, I didn't realise that ValueError is the parent class of UnknownAtomError. I'll insert [an init method](https://www.programiz.com/python-programming/user-defined-exception) to the class with the explicit inheritance with super(), just for completeness.

deeprank_gnn/models/contact.py
```diff
@@ -63,15 +63,19 @@ def residue2(self) -> Residue:
...
+        distance: float,
+        electrostatic_potential: float,
+        vanderwaals_potential: float,
+    ):
 
         self._atom1 = atom1
         self._atom2 = atom2
 
-        Pair.__init__(self, atom1, atom2)
+        super().__init__(atom1, atom2)
```
gcroci2, line 78: I modified the class like this @cbaakman

deeprank_gnn/models/grid.py
```diff
@@ -5,31 +5,39 @@
...
-    """ This holds the value of either one of 4 grid mapping methods.
-        A mapping method determines how feature point values are divided over the grid points.
+    """This holds the value of either one of 4 grid mapping methods.
+    A mapping method determines how feature point values are divided over the grid points.
     """
 
     GAUSSIAN = 1
     FAST_GAUSSIAN = 2
-    BSP_LINE = 3
+    # BSP_LINE = 3
```
cbaakman, line 30: Why is BSP_LINE disabled?

gcroci2: I forgot to uncomment it! I'll do it before merging

PR #63, opened 2022-04-26
fix prospector errors in models/forcefield folder

gcroci2:


deeprank_gnn/models/forcefield/residue.py
```diff
@@ -17,20 +22,22 @@ def matches(self, amino_acid_name: str, atom_names: List[str]) -> bool:
 
         # check the amino acid name
         if self.amino_acid_names != ALL_AMINO_ACIDS:
-            if not any([amino_acid_name == crit_amino_acid_name
-                        for crit_amino_acid_name in self.amino_acid_names]):
+            if not any(
```
gcroci2, line 25: Comprehension inside of 'any' or 'all' is unnecessary. A generator would be sufficient and faster.

cbaakman: I guess, so.

PR #64, opened 2022-04-28
fix prospector errors in example folder

gcroci2:


example/model.py
```diff
@@ -1,141 +0,0 @@
-import numpy as np
```
gcroci2, line 1: We decided to delete this file (2 weeks ago), just confirm

cbaakman: The model.py is not used.
np is unused within the file. I suppose it could go.

PR #68, opened 2022-04-29
Speed up some time-consuming tests

cbaakman:
This makes the total of tests finish in 5 minutes on my local desktop. The changes are:

 - use smaller PDB structures in the tests
 - allow multiprocesssing where possible
 - make sure that some time consuming setup functions are not called repeatedly


gcroci2:
Great! Let's wait for the checks, then we can merge. Hopefully the tests will be faster also in GitHub workflow.

cbaakman:
> Great! Let's wait for the checks, then we can merge. Hopefully the tests will be faster also in GitHub workflow.

The checks just completed successfully. I'll merge.

.github/workflows/ci-pipeline.yml
```diff
@@ -44,11 +44,11 @@ jobs:
       run: |
         # install reduce from git
         git clone https://github.com/rlabduke/reduce
```
gcroci2, line 46: reduce should be clone in ~/src/ folder, otherwise `cmake ~/src/reduce` is not going to work I guess

PR #70, opened 2022-05-02
Update QueryDataset class adding export

FarzanehParizi:
Added a function to QueryDataset class to export the DataSet of Queries to a pickle file.


gcroci2:
The PR is still in progress, right? @FarzanehParizi

PR #71, opened 2022-05-02
fix prospector errors manually in tests folder

gcroci2:


tests/domain/test_amino_acid_d.py
```diff
@@ -2,12 +2,11 @@
```
gcroci2, line 2: For all tests subfolders, I renamed each .py file using the beginning letter of the subfolder and putting it at the end. It's clearer than using __init__.py files, since tests folder is not a package.

tests/feature/test_amino_acid_f.py
```diff
@@ -32,5 +31,5 @@ def test_add_features():
 
     for node in graph.nodes:
         if node.id == variant.residue:  # GLY -> SER
-            node.features[FEATURENAME_SIZEDIFFERENCE] > 0
-            node.features[FEATURENAME_HYDROGENBONDDONORSDIFFERENCE] > 0
+            node.features[FEATURENAME_SIZEDIFFERENCE] > 0 # pylint: disable=pointless-statement
```
gcroci2, line 34: Is this line supposed to do anything?

tests/models/test_structure_m.py
```diff
@@ -0,0 +1,5 @@
+# from deeprank_gnn.models.structure import AtomicElement
```
gcroci2, line 1: Can we delete this file?

tests/domain/test_amino_acid_d.py
cbaakman: OK, I guess the tests will still work this way.

tests/feature/test_amino_acid_f.py
cbaakman: I think that should be preceded by assert. I must have forgotten.

tests/models/test_structure_m.py
cbaakman: yes, let's do that.

tests/feature/test_amino_acid_f.py
gcroci2: perfect I'll add it before merging

PR #80, opened 2022-05-25
Class diagram

gcroci2:
Unified Modeling Language (UML) diagrams generated using Pyreverse, a tool in the [Pylint](https://charlesreid1.com/wiki/Pylint) tool suite.


gcroci2:
Running [this script](https://github.com/DeepRank/deeprank-gnn-2/blob/class_diagram/deeprank_gnn/make_uml.sh) and following the instructions in it I generated several uml in the folder uml/.

[In this folder](https://github.com/DeepRank/deeprank-gnn-2/tree/class_diagram/deeprank_gnn/uml) you can find the main .svg images, and in particular I think that [classes_npl.svg](https://github.com/DeepRank/deeprank-gnn-2/blob/class_diagram/deeprank_gnn/uml/classes_npl.svg) and [packages_npl.svg](https://github.com/DeepRank/deeprank-gnn-2/blob/class_diagram/deeprank_gnn/uml/packages_npl.svg) are the most relevant for our purpose. 

Anyway, running the make_uml.sh script locally (after having installed both pylint and plantuml) you can generate all the files (there can be multiple diagrams for each module), and also the .puml files that I deleted here. You can use them to display interactively the uml, as explained [here](https://chrome.google.com/webstore/detail/plantuml-viewer/legbfeljfbjgfifnkmpoajgpgejojooj).

gcroci2:
Could you please review this PR? @FarzanehParizi

FarzanehParizi:
@gcroci2 @LilySnow 
Better to check the main changes of the final implementation to the first draft of the class diagram before merging. There are many updates and changes compared to the first draft of the class diagram, as the first draft was designed based on the initial requirements.

With an overall check you can see changes for the `Query` Interface and the classes designed to implement it:
e.g.
- The **Query class is not implemented as an Interface**. The interface helps to reduce the code (no implementation is needed in Interface) while providing only the layout for the classes which implement it.
- The PPIInterface class has been changed to two different classes as  `PPIInterfaceAtomicQuery` and `PPIInterfaceResidueQuery`. Although both have mainly the same code copied. They could have been designed with much less code.
   The same for `SingleResidueVariantResidueQuery` and `SingleResidueAtomicQuery`


Screenshot from 2022-08-16 10-36-31

Screenshot from 2022-08-16 10-36-41

Bellow you can see the first draft for this class:

Screenshot from 2022-08-16 21-38-02

gcroci2:
> @gcroci2 @LilySnow Better to check the main changes of the final implementation to the first draft of the class diagram before merging. There are many updates and changes compared to the first draft of the class diagram, as the first draft was designed based on the initial requirements.

The code represented in the class diagrams has been already merged (months ago). But if we think it needs to be changed to be improved, we can of course open new issues. 

> 
> With an overall check you can see changes for the `Query` Interface and the classes designed to implement it: e.g.
> 
> * The **Query class is not implemented as an Interface**. The interface helps to reduce the code (no implementation is needed in Interface) while providing only the layout for the classes which implement it.

Could you be more detailed? Meaning, what should we change in the Query class? For me it's not clear. Maybe it is for @cbaakman ?

> * The PPIInterface class has been changed to two different classes as  `PPIInterfaceAtomicQuery` and `PPIInterfaceResidueQuery`. Although both have mainly the same code copied. They could have been designed with much less code.
>   The same for `SingleResidueVariantResidueQuery` and `SingleResidueAtomicQuery`
> 

I agree that `PPIInterfaceAtomicQuery` and `PPIInterfaceResidueQuery` (and same for `SingleResidueVariantResidueQuery` and `SingleResidueAtomicQuery`) contain basically the same code. But I bet there was a reason for implementing them in this way, right @cbaakman ?

cbaakman:
@gcroci2 I think that the query class cannot be considered an interface anymore, because it has inheritable private methods. I made those so that it's children don't have to implement them. PPIInterfaceAtomicQuery and PPIInterfaceResidueQuery do almost the same, but one builds an atomic graph and the other one builds a residue graph.

gcroci2:
So the questions now are:
- Does the actual Query class, which is not an interface, have limitations? If yes, which ones and why? @FarzanehParizi 
- Regarding PPIInterfaceAtomicQuery and PPIInterfaceResidueQuery: the only real different method between these two classes is `build_graph`, which builds either an atomic graph or a residue graph. This is also the only actual method implemented in the two classes (if you don't consider the dunder methods and get_query_id). Note also that the dunder methods and the get_query_id method do not have much code. I think it makes sense to have two different classes, especially if we think about possible future implementations of atomic-level and residue-level graphs: maybe there will be the need of implementing completely different stuff in the two classes and having the two of them would be handy. But of course, I'm open to discussing it. @LilySnow @FarzanehParizi

gcroci2:
For now, we'll keep the code as it is. In the future, we can take some time to explore the possibility of changing some of the internals, comparing the code to its starting design.

Issue #74, opened 2022-05-09
test_preprocess.py fails on Mac M1 with TypeError: cannot pickle 'module' object


gcroci2:
If I run locally test_preprocess.py on my Mac, it fails:
`FAILED tests/test_preprocess.py::test_preprocess - TypeError: cannot pickle 'module' object`. Maybe issue #56 could solve it, but we need to further investigate it.

cbaakman:
Do you have a stack trace?

gcroci2:
> Do you have a stack trace?

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/Users/giuliacrocioni/Desktop/eScience/projects_2022/3D-vac/3D-Vac/notebooks/1.0GC-pytorch_exploration.ipynb Cell 2' in <cell line: 5>()
      1 metrics_output_directory = "./metrics"
      2 metrics_exporters = [OutputExporter(metrics_output_directory),
      3                      ScatterPlotExporter(metrics_output_directory, 5)]
----> 5 nn = NeuralNet(database, GINet,
      6                node_feature=['type', 'polarity', 'bsa',
      7                              'depth', 'hse', 'ic', 'pssm'],
      8                edge_feature=['dist'],
      9                target='irmsd',
     10                index=range(400),
     11                batch_size=64,
     12                percent=[0.8, 0.2],
     13                metrics_exporters=metrics_exporters)
     15 nn.train(nepoch=50, validate=False)

File ~/Desktop/eScience/projects_2022/3D-vac/deeprank-gnn-2/deeprank_gnn/NeuralNet.py:116, in NeuralNet.__init__(self, database, Net, node_feature, edge_feature, target, lr, batch_size, percent, database_eval, index, class_weights, task, classes, threshold, pretrained_model, shuffle, cluster_nodes, transform_sigmoid, metrics_exporters)
    114         print("the threshold for accuracy computation is set to 0.3")
    115         self.threshold = 0.3
--> 116     self.load_model(database, Net, database_eval)
    118 else:
    119     self.load_params(pretrained_model)

File ~/Desktop/eScience/projects_2022/3D-vac/deeprank-gnn-2/deeprank_gnn/NeuralNet.py:197, in NeuralNet.load_model(self, database, Net, database_eval)
    193 train_dataset, valid_dataset = DivideDataSet(
    194     dataset, percent=self.percent)
    196 # dataloader
--> 197 self.train_loader = DataLoader(
[...]

PR #84, opened 2022-05-30
Fix the preprocessing

cbaakman:
Two changes to the way queries are preprocessed:

- use a python pool for multiprocessing
- do not use pickle to serialize feature modules, pass on the paths instead.


gcroci2:
Fixes #74 and #56. Great, many thanks!

PR #85, opened 2022-05-30
add hdf5dataset as input

gcroci2:
When we use NeuralNet class, the first parameter is the path of the .hdf5 files, and this works well if we don't need to modify anything about them. Here I modify it so that you can pass the "customized" HDF5DataSet object to the NeuralNet class, according to the needs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,15 +191,15 @@ class CustomNet(torch.nn.Module):
 
 
 device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
-nn = NeuralNet(database, CustomNet,
-               node_feature=['type', 'polarity', 'bsa',
-                             'depth', 'hse', 'ic', 'pssm'],
-               edge_feature=['dist'],
-               target='irmsd',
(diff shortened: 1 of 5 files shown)


cbaakman (approved):
It seems OK to do this. The tests still work also.

gcroci2:
> Nice! I really like the idea of separating the data + feature selection and then passing the dataset to the neural network. A possible improvement (but you can do this in a separate issue + PR) is to also separate the data splitting from the neural network training. Then be able to just pass a train and validation set to the neural network (like all major ML and DL frameworks do).
> 
> One note: Is anyone using the repository in one of their projects and will they experience problems if we merge this? (It is quite a major API change). We should start having proper semantic versioning and releases! 😉

I totally agree for the semantic versioning and releases. We're going to announce the first release at tomorrow's daily meeting and we can also ask about eventual problems for this major change. For today I can leave the PR open, just to be sure. 

I also agree on separating data splitting from the neural network training, and I have opened an issue about it (#88).

PR #87, opened 2022-05-31
add a transform_sigmoid unit test

cbaakman:
Run the neural network with transform_sigmoid set to true.

### DIFF
--- a/deeprank_gnn/NeuralNet.py
+++ b/deeprank_gnn/NeuralNet.py
@@ -37,7 +37,7 @@ class NeuralNet():
                  pretrained_model=None,
                  shuffle=True, # pylint: disable=unused-argument
                  cluster_nodes='mcl', # pylint: disable=unused-argument
-                 transform_sigmoid=False, # pylint: disable=unused-argument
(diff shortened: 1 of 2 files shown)


gcroci2:
Before merging, I'd like to add the docstring for transform_sigmoid. I read from the code that it's used only if the task is reg (regression), and it maps the predicted output to the [0,1] domain, but why (or when) would it be necessary? @cbaakman Maybe also @DTRademaker has an answer to this

svenvanderburg:
Hmmm, not sure. I guess it is sometimes useful to map from a [-inf, inf] domain to a [0, 1] or [-1, 1] domain. Basically forcing the values between a certain range.

See below: the x-axis will be the original regressor output, the y-axis will be the new output.
image

I can only think of bad reasons for doing this:
* Your target values are between 0 and 1 and you want to force predictions to be also in this range. But then it would make more sense to use a clip function
* You want to transform your regression output to a binary classification output. But this is not reliable, you just need to train a classifier on binary target data.
But maybe there are good reasons that I don't know of?

PR #102, opened 2022-06-17
fix pip installation

gcroci2:


gcroci2:
I tried to install pytorch and pytorch geometric using pip, modifying the `setup.py` file and linking the wheels as suggested [here](https://stackoverflow.com/questions/66738473/installing-pytorch-with-cuda-in-setup-py) and [here](https://stackoverflow.com/questions/59122308/heroku-slug-size-too-large-after-installing-pytorch/59122860#59122860), but it fails. 

Is it okay to leave their installation to the user (see the [readme](https://github.com/DeepRank/deeprank-gnn-2/tree/fix_pip_installation#:~:text=Dependencies,in%20the%20README))? Also in Deeprank-GNN they did the same (see [here](https://github.com/DeepRank/Deeprank-GNN#:~:text=Installation,https%3A//pytorch.org/)).

PR #109, opened 2022-06-22
Improve continuous workflow and package installation

gcroci2:


gcroci2:
I am following [this](https://packaging.python.org/en/latest/tutorials/packaging-projects/) guide.

Some notes:
- I've added the .toml file for the metadata and I uploaded the deeprankcore package on PyPI, but then the installation was giving errors and so I canceled the project on PyPI.
- Since I've already used deeprankcore as project name on PyPI, now we need to increase the package version in order to be able to upload it with the same name. In this PR everything is already set to an increased package version (0.1.1, the first one was 0.1.0), so when merged we just need to do the release of v0.1.1 on Github to be consistent. Then we can try again to upload the package on PyPI.
- The CI workflow now takes forever, and this is a problem. That's why I want to solve this before merging and reuploading the package on PyPI. I think it's because now the metadata are taken from pyproject.toml instead of setup.py, but the settings seem the same to me in both files. Also I've read that in order to have the package installable with the edit mode, it's good to keep setup.py anyway and it shouldn't cause any problems.

deeprankcore/__init__.py
```diff
@@ -0,0 +1 @@
+"0.1.1"
```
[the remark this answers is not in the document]
svenvanderburg: Maybe this is the default for `bumpver`? You can change it I think using `version_pattern`.

.zenodo.json
```diff
@@ -1,46 +0,0 @@
-{
```
svenvanderburg, line 1: Why did you remove this?

pyproject.toml
```diff
@@ -0,0 +1,71 @@
...
+description = "deeprank-core allows to train graph neural networks to classify protein-protein interface with a greater flexibility for the user."
+readme = "README.md"
+authors = [{ name = "Giulia Crocioni", email = "g.crocioni@esciencecenter.nl" }]
+license = { file = "LICENSE" }
+classifiers = [
+    "License :: OSI Approved :: Apache Software License",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+]
+keywords = ["graph neural network", "protein-protein interface", "pytorch"]
```
svenvanderburg, line 19: This information is all also in `setup.py`... It will be hard to maintain both.

pyproject.toml
```diff
@@ -0,0 +1,71 @@
...
+[build-system]
+requires      = ["setuptools>=61.0.0", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "deeprankcore"
+version = "0.1.1"
+description = "deeprank-core allows to train graph neural networks to classify protein-protein interface with a greater flexibility for the user."
+readme = "README.md"
+authors = [{ name = "Giulia Crocioni", email = "g.crocioni@esciencecenter.nl" }]
```
svenvanderburg, line 12: Add more authors (but they are also in setup.py...)

svenvanderburg (commented):
Cool! 
I added some comments that I think need to be resolved before merging. Feel free to merge afterwards. Let me know if you have questions.

gcroci2:
I added a separated check for prospector, and both checks now pass :). Now one of us can do the release on PyPi, [either manually using twine or configuring GitHub Actions](https://guide.esciencecenter.nl/#/best_practices/language_guides/python?id=packaging-and-distributing-your-package) @svenvanderburg.

PR #125, opened 2022-06-28
adds the transform option to deeprank

cbaakman:
It is tested that this new code doesn't result in any error, when ran.

### DIFF
--- a/deeprankcore/DataSet.py
+++ b/deeprankcore/DataSet.py
@@ -458,12 +458,9 @@ class HDF5DataSet(Dataset):
 
                 ops = [">", "<", "=="]
                 new_cond_vals = cond_vals
-                for op in ops:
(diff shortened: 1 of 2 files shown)


deeprankcore/DataSet.py
```diff
@@ -415,9 +420,8 @@ def create_index_molecules(self):
                     if self.filter(fh5[k]):
                         self.index_complexes += [(fdata, k)]
                 fh5.close()
-            except Exception as inst:
-                print("\t\t--> Ignore File : " + fdata)
-                print(inst)
+            except:
+                _log.exception(f"on {fdata}")
```
gcroci2, line 424: Prospector fails: pylint: `bare-except / No exception type(s) specified`. Please modify line 423 to `except Exception:`

gcroci2 (changes requested):

PR #130, opened 2022-07-01
make hydrogen addition optional

cbaakman:
Hydrogens can contribute to electrostatic forces. However, adding them to structures can be time consuming.
This change makes it optional.


gcroci2 (approved):
Nice useful addition. Feel free to merge :)

PR #131, opened 2022-07-01
makes the model path settable by the user

cbaakman:
This change allows the user to specify the filename to store the neural network model under.


gcroci2 (approved):
Feel free to merge :)

PR #142, opened 2022-08-05
makes dataset use cuda for data tensors

cbaakman:
The dataset class will load the data from the hdf5 file to the cuda device. (if available) This should speed up the training process.


gcroci2 (commented):
Great! I merged :)

Issue #95, opened 2022-06-07
Add integration tests


svenvanderburg:
We don't have integration tests for the whole pipeline (data preprocessing - training a model - evaluation). Therefore it can happen that the preprocessing part of the code doesn't match the training part of the code, see #89 . To prevent this we need integration test(s).

PR #162, opened 2022-08-07
Add integration test

gcroci2:
This PR adds an integration test for the whole pipeline (data preprocessing - training a model - evaluation), see #95

### DIFF
--- a/deeprankcore/preprocess.py
+++ b/deeprankcore/preprocess.py
@@ -35,9 +35,15 @@ def preprocess(feature_modules: List, queries: List[Query],
 
     """
     Args:
-        feature_modules: the feature modules used to generate features, each must implement the add_features function
-        queries: all the queri objects that have to be preprocessed
-        prefix: prefix for the output files, ./preprocessed-data- by default
+        feature_modules: list of feature modules used to generate features.
+        Each feature module must implement the add_features function, and
+        features modules can be found (or should be place in case of custom made feature)
+        in deeprankcore.feature folder.
+
+        queries: all the queri objects that have to be preprocessed.
+
+        prefix: prefix for the output files, ./preprocessed-data- by default.
+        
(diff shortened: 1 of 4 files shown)


deeprankcore/preprocess.py
```diff
@@ -35,9 +35,15 @@ def preprocess(feature_modules: List, queries: List[Query],
 
     """
     Args:
-        feature_modules: the feature modules used to generate features, each must implement the add_features function
-        queries: all the queri objects that have to be preprocessed
-        prefix: prefix for the output files, ./preprocessed-data- by default
+        feature_modules: list of feature modules used to generate features.
+        Each feature module must implement the add_features function, and
+        features modules can be found (or should be place in case of custom made feature)
```
cbaakman, line 40: Maybe this is a chance to fix the typos.

cbaakman:
The test still fails. Can't this test use the 'get_all_scores' function?

gcroci2:
> The test still fails. Can't this test use the 'get_all_scores' function?

First let's fix and merge your PR #164, then I'll modify this one by inserting get_all_scores (after merging this branch with main).

cbaakman:
> > The test still fails. Can't this test use the 'get_all_scores' function?
> 
> First let's fix and merge your PR #164, then I'll modify this one by inserting get_all_scores (after merging this branch with main).

#164 is merged now.

gcroci2:
I added the targets but test_integration.py is still failing. Any idea about the new AssertionError? @cbaakman

cbaakman:
> I added the targets but test_integration.py is still failing. Any idea about the new AssertionError? @cbaakman

The error happens here:
https://github.com/DeepRank/deeprank-core/blob/98728509a47d88c4dd31069908d5220b39568460/tests/test_integration.py#L92

I'm not surprised that this fails. A newly created temporary directory is supposed to be empty. Why are you testing for this?

PR #164, opened 2022-08-09
add documentation on how to add target values

cbaakman:


gcroci2 (changes requested):
Great! Much clearer now. But the tests are failing, feel free to merge once they pass.

cbaakman:
> Great! Much clearer now. But the tests are failing, feel free to merge once they pass.

Indeed! I see the tests are failing. But why is a mystery to me. The fails happen in classes that I didn't even change in this PR. Also, the tests don't fail for me locally. Any suggestions?

gcroci2:
> > Great! Much clearer now. But the tests are failing, feel free to merge once they pass.
> 
> Indeed! I see the tests are failing. But why is a mystery to me. The fails happen in classes that I didn't even change in this PR. Also, the tests don't fail for me locally. Any suggestions?

Found. Scikit-learn released the new 1.1.2 version, and the GitHub workflow is installing this newest version of the package. I bet that you have scikit-learn 1.0.2 on your local machine, and that's why tests/test_graph.py is not failing there. As you can read [here](https://scikit-learn.org/stable/whats_new/v1.1.html#:~:text=random%20sampling%20procedures.-,Fix,10805%20by%20Mathias%20Andersen%20and%20%2323471%20by%20Meekail%20Zain.,-Changelog), [manifold.TSNE](https://scikit-learn.org/stable/modules/generated/sklearn.manifold.TSNE.html#sklearn.manifold.TSNE) now throws a ValueError when fit with perplexity>=n_samples to ensure mathematical correctness of the algorithm.  This is indeed the error that we get on the CI workflow (see [here](https://github.com/DeepRank/deeprank-core/runs/7748362057?check_suite_focus=true#step:4:127:~:text=E%20%20%20%20%20%20%20%20%20%20%20ValueError%3A%20perplexity%20must%20be%20less%20than%20n_samples)). So this test is wrong, and we were not noticing it because of a bug in scikit-learn package. 

The best would be to fix the test, but if requires much effort let me know, I'll merge this PR and then we'll solve this other issue.

cbaakman:
> The best would be to fix the test, but if requires much effort let me know, I'll merge this PR and then we'll solve this other issue.

I've set perplexity to the lowest recommended value: 5.0 . This passes the test, but I've no idea if this is still allright.

PR #179, opened 2022-09-14
Remove rename params

DaniBodor:
remove threshold param from NeuralNet class and make it required in Metrics class


gcroci2 (approved):
Great work! Let's remove the threshold parameter. Feel free to merge.

PR #186, opened 2022-09-16
HDF5DataSet subset (formerly index)

DaniBodor:
renamed and fixed issues regarding this parameter


docs/quickstart.md
```diff
@@ -53,7 +53,7 @@ metrics_exporters = [OutputExporter(metrics_output_directory),
 dataset = HDF5DataSet(
     root = "./",
     database = database,
-    index = range(400),
+    subset = range(400),
```
gcroci2, line 56: We should insert keys here, but let's just remove `subset` from the quickstart file for now.

gcroci2 (commented):
Great! Another parameter's usage clarified.
I would suggest to also add a unit test in tests/test_dataset.py file since we are never testing the `subset` parameter when it's not None. You could define a test in which you open the hdf5 file, save the keys in a variable, select only n keys and pass them to `subset`. Then assert if the number of the keys you selected equals the dataset length.

DaniBodor:
@gcroci2 , does it automatically update my pull request with the new stuff or do I need to do something?

gcroci2:
> @gcroci2 , does it automatically update my pull request with the new stuff or do I need to do something?

It does it automatically when you commit to the branch of the PR :)

gcroci2:
You can also see your last commit above "added unit test for subset parameter" @DaniBodor

DaniBodor:
OK!
Was I supposed to have created a separate branch and PR from this one, so that you can quickly check what has changed?
I guess because it was only few small changes it's fine, but for bigger changes would that make sense?

gcroci2:
> OK! Was I supposed to have created a separate branch and PR from this one, so that you can quickly check what has changed? I guess because it was only few small changes it's fine, but for bigger changes would that make sense?

In general, if the changes are related to the same issue/PR it's fine to commit them in the same branch/PR, like you did.

PR #187, opened 2022-09-16
Percent param

DaniBodor:
I changed `percent` to `train_size`, following nomenclature of [sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html) and removed `shuffle` parameter.

I changed it such that floats or ints can be used as input parameter and defaulted it to 0.75, as discussed in issue 183.

Also added checks to make sure that the value given makes sense and throwing errors in case it doesn't.

Updated explanations of functions accordingly and added some additional info.

Are any additional unit tests required?


deeprankcore/DataSet.py
```diff
@@ -54,36 +54,58 @@ def save_hdf5_keys(
...
+        
+    # negative values means it is the number of datapoints to NOT include in training data
+    if n_train < 0:
+        n_train = full_size + n_train # adding negative number
+    
+    # raise exceptions if no training data or if more than 100% training data
+    if n_train > full_size:
+        raise ValueError ("invalid train_size: train_size too large. \n\t" +
+            f"train_size must be a float between -1 and 1 OR an int with max absolute value of len(dataset) ({full_size})")
+    if n_train <= 0:
```
gcroci2, line 92: I would suggest removing the possibility to insert negative values. These are the constraints for `train_size` from sklearn `train_test_split` and I would stick with those: 
- Float: check that is between 0.0 and 1.0
- Int: check that is positive and <= total number of samples
- None: defaults to 0.75
They are basically already implemented, I would just add a check for float and a check for int, removing the negative value option

deeprankcore/NeuralNet.py
```diff
@@ -41,16 +41,17 @@ def __init__(self, # pylint: disable=too-many-arguments, too-many-locals
...
             dataset_test (HDF5DataSet object, optional): independent evaluation set. Defaults to None.
             lr (float, optional): learning rate. Defaults to 0.01.
             weight_decay (float, optional): weight decay (L2 penalty). Weight decay is 
                     fundamental for GNNs, otherwise, parameters can become too big and
                     the gradient may explode. Defaults to 1e-05.
             batch_size (int, optional): defaults to 32.
-            percent (list, optional): divides the input dataset into a training and an evaluation set.
-                    Defaults to [0.8, 0.2].
+            train_size (float or int, optional): fraction of dataset (if float) or number of datapoints (if int) to use for training. 
+                Negative values will set size of the validation dataset instead.
```
gcroci2, line 53: Remember to remove the comment about negative values if you change _DivideDataSet

gcroci2 (commented):
Nice edits :) I added a minor comment about negative values in the review. I would also like to add a unit test for this parameter. We could test only _DivideDataSet, giving it as input an HDF5DataSet object and asserting the number of data points after the splitting.

deeprankcore/DataSet.py
DaniBodor: The difference though is that in train_test_split you can set the train_size, test_size, or both (can be less than 100%). In that way you can just set the val_size directly. I figured to keep the number of parameters low, but still allow for the same functionality, it's nice to be able to use negative values.
I realize that it's a bit of overkill to allow all of these options, as they determine each other and there are different ways to get it done, but it does allow for different ways of thinking to work. For example, I would think more in terms of how much data I am willing to sacrifice as validation data, rather than in terms of how much I want to keep as training data. That's why I like being able to set it directly (also in train_test_split, it is primarily the test data that is set and train data that follows).

Also, in a way it makes sense that negative values determine the amount of data to NOT use for training, where positive values determine the amount of data to USE for training.

gcroci2: > The difference though is that in train_test_split you can set the train_size, test_size, or both (can be less than 100%). In that way you can just set the val_size directly. I figured to keep the number of parameters low, but still allow for the same functionality, it's nice to be able to use negative values. I realize that it's a bit of overkill to allow all of these options, as they determine each other and there are different ways to get it done, but it does allow for different ways of thinking to work. For example, I would think more in terms of how much data I am willing to sacrifice as validation data, rather than in terms of how much I want to keep as training data. That's why I like being able to set it directly (also in train_test_split, it is primarily the test data that is set and train data that follows).
> 
> Also, in a way it makes sense that negative values determine the amount of data to NOT use for training, where positive values determine the amount of data to USE for training.

Also consider that this functionality is not supposed to be used a lot, it's just something that for now we're leaving there because it was already there and it could (very rarely) be handy. In the majority of cases, users are going to do the split before instantiating NeuralNet (and they do not want to randomly split data after shuffling; this is often done in ML tutorials, but in real use cases you want to have a clear distinction between which groups are in train and which ones are in validation, and this is especially true in our project domain). If you prefer we can have `test_size` instead of `train_size`, this would be fine for me.

PR #190, opened 2022-09-20
Auto test all features

DaniBodor:
- created generic function in test_preprocess.py instead of having 2 copies of same function
- in 2nd test: read all modules from deeprankcore.feature using importlib and test those

### DIFF
--- a/tests/test_preprocess.py
+++ b/tests/test_preprocess.py
@@ -13,10 +13,10 @@ from typing import List
 
 def preprocess_tester(feature_modules: List):
     """
-    Generic function to test preprocess for either single or all feature types.
+    Generic function to test preprocessing several PDB files into their feature representation HDF5 file.
 
     Args:
-        feature_modules: list of feature modules to be tested
+        feature_modules: list of feature modules (from .deeprankcore.feature) to be passed to preprocess
     """
 
     output_directory = mkdtemp()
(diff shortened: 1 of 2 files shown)


tests/test_preprocess.py
```diff
@@ -1,24 +1,24 @@
 from tempfile import mkdtemp
 from shutil import rmtree
 import os
+from xml.dom.minicompat import defproperty
```
DaniBodor, line 4: I have no recollection at all of adding this.
Just remove it, it is not used anywhere

DaniBodor:
not doing what I thought it was doing. I'll create a new PR when it does

DaniBodor:
> not doing what I thought it was doing. I'll create a new PR when it does

I fixed the problem

tests/test_preprocess.py
```diff
@@ -50,43 +54,28 @@ def test_preprocess_one_feature():
...
+def test_preprocess_single_feature():
     """
-    Tests preprocessing several PDB files into their features representation HDF5 file.
+    Tests preprocessing for single feature.
     """
 
-    output_directory = mkdtemp()
-
-    prefix = os.path.join(output_directory, "test-preprocess")
+    imp = importlib.import_module(('deeprankcore.feature.sasa'))    # linting doesn't like: from deeprankcore.feature import sasa below top_level
```
gcroci2, line 63: You could import sasa from outside the function (very top of the script where are the other modules imports) and then just do `preprocess_tester([sasa])`

gcroci2 (approved):

tests/test_preprocess.py
DaniBodor: I know, but preferred it this way because it's consistent with what I do in the 2nd test. I forgot to remove that comment, it was a note to self more than anything else.

gcroci2: Okay, both are fine :)

Issue #154, opened 2022-08-05
add the possibility to set the optimizer during training phase


gcroci2:
Adam is great as default, but then it's nice to have the possibility to choose

gcroci2:
Blocked by #172

PR #193, opened 2022-09-21
Align optimizer

gcroci2:
In NeuralNet class, the optimizer used during training is Adam and can't be set by the user.

With this PR the user can set the optimizer using `configure_optimizer` method of NeuralNet class. 
The default is Adam, and I changed the default learning rate from 0.01 to 0.001 (the default one, see [PyTorch documentation](https://pytorch.org/docs/stable/generated/torch.optim.Adam.html)).

Refers to issue #154


cbaakman (approved):
It seems OK to have the optimizer configuration in a different class method.

PR #195, opened 2022-09-26
Atom feature

DaniBodor:
Created a module feature.atom.py with features for atom nodes.
Data is loaded from models.structure.Atom and feature.atomic_contact.add_features_for_atoms.
Also created tests/features/test_atom to check that this is working.


deeprankcore/feature/atom.py
```diff
@@ -0,0 +1,53 @@
...
+# these will ultimately be moved to deeprankcore.domain.feature.py and imported in this file.
+# for now, I am keeping them here to avoid merge conflicts with the feature renaming branch.
+
+# from models.structure.Atom
+FEATURE_NODE_ATOMICELEMENT = "atomic_element" # AtomicElement object
+FEATURE_NODE_PDBOCCUPANCY = "pdb_occupancy" # float(0 < x < 1)
+
+# from feature.atomic_contact.add_features_for_atoms
+FEATURE_NODE_ATOMICCHARGE = "atomic_charge" # float
+FEATURE_NODE_VANDERWAALS = "vanderwaals" # ?
```
DaniBodor, line 21: These will move to deeprankcore.domain.feature.py.
I am not doing this yet because I am editing that file in another branch and it would probably create a merge conflict.

DaniBodor:
Is there a way to set coveralls to not "fail" if the decrease is below some threshold?

deeprankcore/feature/atom.py
```diff
@@ -0,0 +1,46 @@
...
+    ):
+
+    for node in graph.nodes:
+        if isinstance(node.id, Atom):   # I think this is necessary, otherwise it'll complain if it's an amino acid graph, right?
+            atom = node.id
+
+            node.features[FEATURE_NODE_ATOMICELEMENT] = atom.element.onehot
+            node.features[FEATURE_NODE_PDBOCCUPANCY] = atom.occupancy
+        
+            try:    # I copied the try/except structure from feature.atomic_contact.add_features_for_atoms; not sure if it's actually necessary.
```
gcroci2, line 37: This part is handled by deeprankcore.feature.atomic_contact, so there is no need of creating the same features here. The logic there is also a bit different because it takes into account only atoms in contact. Let's consider this module only for simple single-atoms features, such as atomic type.

deeprankcore/feature/atom.py
```diff
@@ -0,0 +1,46 @@
...
+    graph: Graph, 
+    *args, **kwargs
+    ):
+
+    for node in graph.nodes:
+        if isinstance(node.id, Atom):   # I think this is necessary, otherwise it'll complain if it's an amino acid graph, right?
+            atom = node.id
+
+            node.features[FEATURE_NODE_ATOMICELEMENT] = atom.element.onehot
+            node.features[FEATURE_NODE_PDBOCCUPANCY] = atom.occupancy
```
gcroci2, line 35: I'm not sure if this is a feature that may be useful (I checked for some tests data and it's always 1.0, for example), but I asked Coos and I'll let you know.

deeprankcore/feature/atom.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+_log = logging.getLogger(__name__)
+
+
+# these will ultimately be moved to deeprankcore.domain.feature.py and imported in this file.
+# for now, I am keeping them here to avoid merge conflicts with the feature renaming branch.
+
+# from models.structure.Atom
+FEATURE_NODE_ATOMICELEMENT = "atomic_element" # AtomicElement object
```
gcroci2, line 16: I'd call it "atom_type" as agreed with Daniel (I think that it's not really the atomic element, indeed it can have only 6 values, much less than the actual number of existing atoms; I asked clarifications for this to Daniel and I'll let you know)

gcroci2 (changes requested):
Well done! In general I think that this preprocessing part is the trickiest of the package and it's good for both of us to start understanding it better. Please remove the last part of deeprankcore.feature.atom (see comment) and add a test for the feature you added (for now only for "atomic_element"/"atom_type", see comment)

DaniBodor:
> Please remove the last part of deeprankcore.feature.atom (see comment) 
I'm not sure which comment you mean, but after our discussion this morning I think I should leave it all there, right?

> and add a test for the feature you added (for now only for "atomic_element"/"atom_type", see comment)
It is not clear to me what should be in this test. Can you help me with this?

gcroci2:
> I'm not sure which comment you mean, but after our discussion this morning I think I should leave it all there, right?

Yes exactly! I didn't update the comment

> It is not clear to me what should be in this test. Can you help me with this?

See scripts in tests.feature as examples for creating a test for the new feature module. If still is unclear, we can do it together

gcroci2 (approved):

DaniBodor:
Regarding the pdb occupancy feature: I guess it gives some level of confidence of the accuracy of the position of the atom in question. This could in theory be a consequential feature. For example: maybe the network will learn to give lower accuracy atoms less weight.
I can remove it if you think it's not relevant, but my impression was always that we should let the network decide/learn what is and what isn't a relevant parameter as opposed to try to use human reasoning for it.
What do you think @gcroci2 and @cbaakman ?

DaniBodor:
@gcroci2, I added an test_atom module. I just had it check that ATOMCHARGE and PDBOCCUPANCY exists for now (easiest because it's numerical and I can do an isnan test). Do I need to check for each feature or is it ok like this? 
I saw that in test_amino_acid, also only few of the features were explicitly checked, so guess that this should be enough to see that it is generating features.

gcroci2 (approved):

gcroci2 (approved):

gcroci2:
> @gcroci2, I added an test_atom module. I just had it check that ATOMCHARGE and PDBOCCUPANCY exists for now (easiest because it's numerical and I can do an isnan test). Do I need to check for each feature or is it ok like this? I saw that in test_amino_acid, also only few of the features were explicitly checked, so guess that this should be enough to see that it is generating features.

It's fine like this for now :)

PR #196, opened 2022-09-26
NeuralNet class alignment to high-level frameworks

gcroci2:
Main edits

**General**
- Improved documentation in Trainer.py and DataSet.py
- Added `_` in front of methods meant to be only internal to the classes, so not supposed to be called by the user or somewhere outside the classes, in Trainer.py _and_ DataSet.py. 

**DataSet.py**
- Inserted a check for the target Dataset to be numerical since we do not support the categorical targets (for our experiments we have numerical target values so we don't care for now, but it's a nice addition for the future - see issue #197).

**NeuralNet.py (Trainer.py)**
- Modified NeuralNet class and module to Trainer
- Moved `_PreCluster` and `_DivideDataSet` functions to Trainer.py module since they are used only in there. I put `_PreCluster` as `Trainer` method and I left `_DivideDataSet` as a function for now. In the future, we may want to move `_PreCluster` and its calls to HDF5DataSet class (I think it makes more sense to have it there), and `_DivideDataSet` could be made a Trainer method.
- Moved `classes_to_idx` to `Trainer.format_output` and changed it from attribute to local method variable since it's used only there.
- Removed `Trainer.compute_class_weights` method because it was never used. 
- Removed `Trainer.update_name` method because it was never used. 
- Raise an error if both `dataset_val` and `val_size` are set. Only one between them should be set. 
- If there is no `pretrained_model`, raise an error if either:
  - there is no `dataset_train`
  - there is no `Net`
- in `Trainer._load_pretrained_model` method, create self.test_loader _after_ having called `_PreCluster`.
- in `Trainer.train` method, raise an error if `validate` is True and there is no validation loader. 
- in `Trainer.test` method:
  - call `_PreCluster` only if self.cluster_nodes is in ('mcl', 'louvain')
  - raise an error if `dataset_test` provided to `test()` is None **and** `test_loader` attribute is None (meaning no dataset test was provided neither when Trainer was initialized).

### DIFF
--- a/deeprankcore/Trainer.py
+++ b/deeprankcore/Trainer.py
@@ -16,7 +16,7 @@ from deeprankcore.DataSet import _DivideDataSet, PreCluster
 _log = logging.getLogger(__name__)
 
 
-class NeuralNet():
+class Trainer():
 
     def __init__(self, # pylint: disable=too-many-arguments
                  Net,
@@ -37,7 +37,7 @@ class NeuralNet():
         Args:
             Net (function, required): neural network class (ex. GINet, Foutnet etc.).
                 It should subclass torch.nn.Module, and it shouldn't be specific to regression or classification
-                in terms of output shape (NeuralNet class takes care of formatting the output shape according to the task).
+                in terms of output shape (Trainer class takes care of formatting the output shape according to the task).
                 More specifically, in classification task cases, softmax shouldn't be used as the last activation function.
             dataset_train (HDF5DataSet object, required): training set used during training.
             dataset_val (HDF5DataSet object, optional): evaluation set used during training.
@@ -96,7 +96,7 @@ class NeuralNet():
                         "User target detected -> The task argument is required ('class' or 'reg'). \n\t"
                         "Example: \n\t"
                         ""
-                        "model = NeuralNet(dataset, GINet,"
+                        "model = Trainer(dataset, GINet,"
                         "                  target='physiological_assembly',"
                         "                  task='class',"
                         "                  val_size=0.25)")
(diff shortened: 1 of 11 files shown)


deeprankcore/Trainer.py
```diff
@@ -582,49 +592,34 @@ def log_epoch_data(stage, epoch, loss, time):
...
+            # For categorical cross entropy, the target must be a one-dimensional tensor
+            # of class indices with type long and the output should have raw, unnormalized values
+            classes_to_idx = {
+                i: idx for idx, i in enumerate(self.classes)
+            }
 
-            # Sigmoid(x) = 1 / (1 + exp(-x))
-            pred = torch.sigmoid(pred.reshape(-1))
+            target = torch.tensor(
+                [classes_to_idx[int(x)] for x in target]
```
cbaakman, line 606: I know that the classes_to_idx translation table isn't used much, since we always work with numerical targets anyway. However, I do think is should be a property of the run, rather than a temporary value. After all, this information is needed at the end of the run to convert the indexes back into class labels.

deeprankcore/Trainer.py
```diff
@@ -183,10 +210,10 @@ def load_model(self, dataset_train, dataset_val, dataset_test, Net):
         if self.cluster_nodes is not None:
             if self.cluster_nodes in ('mcl', 'louvain'):
```
cbaakman, line 211: Wouldn't it be better to replace such input parameters by enumeration objects?

deeprankcore/Trainer.py
```diff
@@ -272,10 +303,6 @@ def put_model_to_device(self, dataset, Net):
         # classification mode
         elif self.task == "class":
```
cbaakman, line 304: Wouldn't it be a good idea to use enumeration objects here? For CLASS / REGRESSION

deeprankcore/Trainer.py
gcroci2: You're right, I moved it back to `_put_model_to_device` and I made it an attribute of Trainer again.

deeprankcore/Trainer.py
gcroci2: I know enumeration in Python but I'm not getting what you mean. An example for the above code?

PR #199, opened 2022-09-29
Explore target

DaniBodor:
- moved `task` parameter from `NeuralNet` to `HDF5DataSet`, so that it is in the same class as `target`
- as task options, I changed "class" and "reg" to "classif" and "regress", which to me are clearer
- added some linting to NeuralNet class


DaniBodor:
For some reason all tests in test_nn.py and test_integration.py are failing with a message 

> AttributeError: 'NeuralNet' object has no attribute 'task'

I don't understand why this is happening. As far as I see, it doesn't try to assign a `task` to the NeuralNet objects, only to the HDF5DataSet objects.

DaniBodor:
I think this is going to have a bunch of conflicts with #196 (am still in the process of reviewing that).
What is the best way to go forward?

docs/quickstart.md
```diff
@@ -67,7 +67,36 @@ Data can be split in sets implementing custom splits according to the specific a
 Assuming that the training, validation and testing ids have been chosen (keys of the hdf5 file), then the corresponding graphs can be saved in hdf5 files containing only references (external links) to the original one. For example:
 
 ```python
-from deeprankcore.DataSet import save_hdf5_keys
+from deeprankcore.NeuralNet import NeuralNet
```
gcroci2, line 70: I think that `from deeprankcore.DataSet import save_hdf5_keys` should stay and lines from 70 to 99 should be deleted (see below, HDF5DataSet and NeuralNet objects are initialized there)

gcroci2 (changes requested):
Everything is fine, except in the readme we need to add the new task parameter in the HDF5DataSet objects

DaniBodor:
Replaced by @200

PR #200, opened 2022-09-30
task and classes to HDF5DataSet

DaniBodor:
- moved `task` and `classes` parameters from `Trainer` to `HDF5DataSet`, so that they are in the same class as `target`
- as task options, I changed "class" and "reg" to "classif" and "regress", which to me are clearer
- renamed test_nn.py to test_trainer.py to conform to new nomenclature


deeprankcore/DataSet.py
```diff
@@ -118,6 +125,32 @@ def __init__( # pylint: disable=too-many-arguments
...
         self.target = target
+
+        # I made it so it overrules the user setting in case the target is one of the 'known' types.
+        # Is this what we want, or do we prefer to raise an error or show a warning in case it does not match? 
+        if self.target in ["irmsd", "lrmsd", "fnat", "dockQ"]:
+            self.task = "regress"
+        elif self.target in ["bin", "binary", "bin_class", "capri_classes"]:
+            self.task = "classif"
+        else:
+            self.task = task
```
DaniBodor, line 136: I made it so it overrules the user setting in case the target is one of the 'known' types. Is this what we want, or do we prefer to raise an error or show a warning in case it does not match?

DaniBodor: @gcroci2, did you see this?

gcroci2: Yes it's fine :)

PR #201, opened 2022-10-06
Domain overhaul

DaniBodor:
### Update deeprankcore/domain/
- Reorganized domain/features.py and domain/storage.py into submodules for groups, nodefeatures, edgefeatures
- Created a domain/targettypes.py module for default target and task calls
- Renamed features for brevity & clarity (kept old names as comments in code)
- Replaced strings by their feature-variables throughout
- Replaced `HDF5KEY_GRID_XXXX` variables for strings (only used in 1 module and 1 associated unit test)
- Updated README and quickstart to use feature-variables as well
  - Removed `RESTYPE` and `POLARITY` from feature lists in these documents, as the feature modules creating them are not used in the example

### Specific feature stuff
- Moved metafeatures (non-trainable features such as node names, etc) inside main feature groups of hdf5 files as underscored features (e.g. "node_features/_name")
  - features starting with an underscore are ignored by the rest of the pipeline
  - made `POSITION` a metafeature rather than a core feature, because GNNs should not use position information
- Assigned currently unused features
  - `CHARGE` (wasn't previously assigned) + `DIFFCHARGE` (new feature for variant residues)
  - `CHAINID` (wasn't previously assigned); 
    - This is not a trainable feature so I underscored it as "_chain_id" (like "_names" and "_indices"). In theory, this could be removed, as the info is also visible in `NAMES`, but it is currently used as an intermediary in tools/visualization and I can't be bothered to figure out how to get the info there directly.
  - `SAMECHAIN` (used to always be set to 1)
    - Used to be in tools.visualization, line 83: `graph.edges[node1_name, node2_name][edgefeats.SAMECHAIN] = 1.0` without any if statement. Now the `CHAINID` is checked and `SAMECHAIN` set accordingly in feature/atomic_contact.py
  -`SAMERES` (new feature for atom graphs to check whether both atoms are part of same residue; analogous to `SAMECHAIN`)
- Removed obsolete/duplicate features
  - ~~`EDGETYPE`~~ =`SAMECHAIN`
  - ~~`CONSERVATIONDIFFERENCE`~~/~~`PSSMDIFFERENCE`~~ =`DIFFCONVERSATION`
  - ~~`VARIANTCONSERVATION`~~ & ~~`PSSMVARIANT`~~ are also covered by `DIFFCONVERSATION` (in the same way other variant features are also only covered by the difference)

### Unit testing (ALL DONE)
- Tested that HDF5 files are created correctly with this branch
- Added (temporary) jupyter notebook to create HDF5 files
- Recreated the HDF5 files needed for unit testing
- Moved (new) HDF5 files into tests/data/hdf5 instead of tests/hdf5

### Other (ALL DONE)
- Moved AminoAcid and Polarity classes into deeprankcore/models/amino-acid
- Made test_integration faster by decreasing the number of queries and epochs
- Deleted obsolete GraphGenMP.py file (+ its unit test) and the example folder
- Moved CUDA testing into separate test of test_trainer instead of making it always run CUDA if available (#204)

### ~~Update cluster naming to one level shallower and prefixing the clustering-type~~
Decided against this after all; potentially there can be multiple clusters in the same HDF5 file (I think) and nicer organized to have them in subfolders. Also, the code for this is kinda messy and I don't want to screw it up right now. It can be a future UX/UI issue if we want to go this route after all


gcroci2 (approved):
Great!! Much more clearer now :-) I corrected the variants.hdf5 file to have 'binary' as target. Now everything should be fine.

cbaakman (approved):
Looks good. It must have been a lot of work.

gcroci2 (approved):

gcroci2:
I'm going to regenerate tests data and re-run experiments scripts this morning

PR #208, opened 2022-10-18
Adds GPU integration test support

namiyousef:
Fixes https://github.com/namiyousef/colab-utils/issues/12

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,28 +31,15 @@ jobs:
           pytest -v
       - name: Verify that we can build the package
         run: python3 setup.py sdist bdist_wheel
-    
+  
   integration:
     name: Test with GPU
     runs-on: ubuntu-latest
-
-    strategy:
-      fail-fast: false
-      matrix:
-        os: ['ubuntu-latest']
-        python-version: [3.9]
-    
-    
-
     steps:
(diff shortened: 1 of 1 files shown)


namiyousef:
I believe the error results from there not being a Kaggle API key. If you add a valid API key to the project secrets under the name "KAGGLE_API_KEY" it should work

gcroci2:
> I believe the failed GPU integration tests results from there not being a Kaggle API key. If you add a valid API key to the project secrets under the name "KAGGLE_API_KEY" it should work

Thank you very much for investigating this and opening a new PR. I checked and there was a kaggle api key, but to be sure I generated a new one, and it's still failing. Could it fail because of the package that is not installed in the "Test with GPU" job? @namiyousef

namiyousef:
Hi @gcroci2, thanks for looking into this. I believe there was an issue with the api key earlier (hence the json decode error) but it seems like you have fixed it. I looked into your comment and revisited the issue you created here: https://github.com/namiyousef/colab-utils/issues/12. The problem is that the Kaggle runner uses Python 3.7. It doesn't seem like they provide support for other python versions, so this will remain an open issue. 

Thanks for looking into this once again. If I find a better solution for this problem I will let you know.

Closing the MR

Issue #218, opened 2022-11-01
suspected mistake in `get_lennard_jones_potentials` function


DaniBodor:
in features.atomic_contact the `get_lennard_jones_potentials` function ends with

```
    # calculate potentials
    sigmas = 0.5 * (sigmas1 + sigmas2)
    epsilons = np.sqrt(sigmas1 * sigmas2)
    potentials = 4.0 * epsilons * ((sigmas / distances) ** 12 - (sigmas / distances) ** 6)
```

I suspect it should be (3rd line different)

```
    # calculate potentials
    sigmas = 0.5 * (sigmas1 + sigmas2)
    epsilons = np.sqrt(epsilons1 *epsilons2)
    potentials = 4.0 * epsilons * ((sigmas / distances) ** 12 - (sigmas / distances) ** 6)
```

can you confirm, please @cbaakman

cbaakman:
Yes, sorry about the mistake!

PR #220, opened 2022-11-01
Simplified code for edge features

### LINKED ISSUE #207: Explore `add_features_for_residues` function in feature/atomic_contact.py
Currently nested for loops are used to compare each atom of one residue against each atom of the other residue and extract information one comparison at a time.
There is likely a way to get the sum/minimum/etc (whichever is applicable) in one shot, perhaps using some numpy function. 

This is likely to be faster and definitely more pythonic and readable.

DaniBodor:
Massively simplified code for addition of edge features in features.atomic_contact.
Also fixed mistake in LJ potential calculation
Removed old functions from tools.pdb, which I think were intended for or once upon a time used for generating edge features, but is now done in feature.atomic_contact

closes #207 and #218


deeprankcore/domain/forcefield/__init__.py
```diff
@@ -58,17 +59,12 @@ def _find_matching_residue_class(self, residue: Residue):
...
-    def _get_type(self, atom: Atom):
         atom_name = atom.name
 
         if atom.residue.amino_acid is None:
-            raise UnknownAtomError(f"no amino acid for {atom}")
-
-        residue_name = atom.residue.amino_acid.three_letter_code
+            _log.warning(f"no amino acid for {atom}; three letter code set to XXX")
+            residue_name = 'XXX'
+        else: residue_name = atom.residue.amino_acid.three_letter_code
```
DaniBodor, line 68: this error (and below) used to be caught and handled in the code calling this function. I just substituted that warning into here instead.

deeprankcore/domain/forcefield/__init__.py
```diff
@@ -58,17 +59,12 @@ def _find_matching_residue_class(self, residue: Residue):
...
-
-        return self._vanderwaals_parameters[type_]
-
-    def _get_type(self, atom: Atom):
         atom_name = atom.name
 
         if atom.residue.amino_acid is None:
-            raise UnknownAtomError(f"no amino acid for {atom}")
-
-        residue_name = atom.residue.amino_acid.three_letter_code
```
DaniBodor, line 71: this and below UnknownAtomErrors used to be caught in the code that calls the functions. I just substituted the warnings raised there in here.

DaniBodor:
All tests run fine locally, not sure why they fail here

deeprankcore/domain/forcefield/__init__.py
```diff
@@ -58,17 +59,12 @@ def _find_matching_residue_class(self, residue: Residue):
         return None
 
     def get_vanderwaals_parameters(self, atom: Atom):
-        type_ = self._get_type(atom)
```
gcroci2, line 61: Don't we need type_ var?

gcroci2 (approved):

deeprankcore/domain/forcefield/__init__.py
cbaakman: It's used for getting the right vanderwaals parameters from the table file.

gcroci2 (commented):
Seems good, but the build is failing. Please fix it before merging :)

DaniBodor:
I noticed that this code is actually slower than the original in assigning features. I tested this by looping some tests from test_query and checking how long it takes to run.

I realize that this is (probably) because before that calculations (e.g. `get_coulomb_potentials` and `get_lennard_jones_potentials`) were only run once per graph on many atoms, and now they are run once per edge on few atoms at a time. Apparently the new way is slower.

I can look into finding a way to combine what happened before with the simplified code base, or we can revert to how it was or leave it as is. What do you think?

gcroci2:
> I noticed that this code is actually slower than the original in assigning features. I tested this by looping some tests from test_query and checking how long it takes to run.
> 
> I realize that this is (probably) because before that calculations (e.g. `get_coulomb_potentials` and `get_lennard_jones_potentials`) were only run once per graph on many atoms, and now they are run once per edge on few atoms at a time. Apparently the new way is slower.
> 
> I can look into finding a way to combine what happened before with the simplified code base, or we can revert to how it was or leave it as is. What do you think?

I would just revert it. It would have been a nice addition but I wouldn't lose too much time on this

deeprankcore/domain/forcefield/__init__.py
DaniBodor: As far as I could see it was only used internally by the `get_vanderwaals_parameters` function. However, this function basically did nothing else than read the type_ and get the parameters, so instead I merged `get_type` into `get_vanderwaals_parameters`.
I tested that old and new `add_features` functions (from atomic_contact, which uses this) give the exact same result on ~5 test cases, so am fairly confident that it is still being read correctly.

DaniBodor:
> I would just revert it. It would have been a nice addition but I wouldn't lose too much time on this

I resolved it. Didn't take that long and wanted to use some of the fixes from the PR anyway.

gcroci2 (approved):

PR #223, opened 2022-11-03
Revert "deleted obsolete module+test"

DaniBodor:
This reverts commit 5d3ad97afdf60fa5fdcf6512e567b2092d4c9540.


DaniBodor:
this branch is identical to the last stable commit (b103b726fe544771b32ee0885aedd1f692b65079) and still fails. I did a git diff on these two and get nothing.

I suspect that the failed build is due to something not related to the code, as all tests succeed locally. Could it perhaps have to do with the pytorch version or something? Similar to the problems we had on Snellius.

PR #226, opened 2022-11-08
Atomiccontact speed

DaniBodor:
rewrote function to make it faster


DaniBodor:
Function is no longer slow

PR #240, opened 2022-11-11
style: restructure deeprankcore package and subpackages

gcroci2:
Reminder: modifying tools/ folder, cleaning not used files, moving files to more appropriate places. _Done_

To continue: tools/pssm should be merged with models/pssm. However, on first try, this created circular calls between models.pssm and models.structure. _We called them operations/pssmdata and operations/parsing/pssm for now._


deeprankcore/models/operations/parsers/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+from deeprankcore.models.structure.residue import Residue
+from deeprankcore.models.operations.parsers.top import TopParser
+from deeprankcore.models.operations.parsers.patch import PatchParser, PatchActionType
+from deeprankcore.models.operations.parsers.residue import ResidueClassParser
+from deeprankcore.models.operations.parsers.vdwparam import VanderwaalsParam, ParamParser
+
+
+_log = logging.getLogger(__name__)
+
+_forcefield_directory_path = os.path.realpath(os.path.join(os.path.dirname(__file__), '../../../domain/forcefield'))
```
DaniBodor, line 13: is there a cleaner/prettier way to define this path?

gcroci2: I looked a bit but being the file nested in domain/forcefield other ways are very similar to this. Since we removed one level for operations (deleting models) we have one less `..` now

DaniBodor: It's more that I don't like the fact that if we change something in the package structure we need to change this manually. Was wondering whether there is some nice pythonic way to read the folder directly, similar to an import. But probably not worthwhile investigating it if you don't have a ready to go solution.

gcroci2:
In uniforming deeprankcore nomenclature, I used only _deeprankcore_ (name of the package on PyPi) and _DeeprankCore_, even if the repo is called deeprank-core (it should be _deeprankcore_ for consistency). We plan to change the repo's name in deeprank after the generalization project will be over.

deeprankcore/domain/edgefeatures.py
```diff
@@ -1,3 +1,10 @@
+## main group
+EDGE = "edge_features"
+
+## metafeatures
+NAME = "_name"
+INDEX = "_index"
```
cbaakman, line 6: What are the main group and metafeatures for?

cbaakman (commented):
Why are atom and sasa features removed?

gcroci2:
> Why are atom and sasa features removed?

We moved atom in components.py, and sasa in surfacearea.py

deeprankcore/domain/edgefeatures.py
gcroci2: main group indicates the hdf5 Group name, containing the hdf5 Datasets whose names are listed below. In edges case it's called `edge_features`, and we have the same structure for nodes. Metafeatures are the Datasets' names used for identifying the edges that are not used as features for training the networks.

deeprankcore/features/components.py
```diff
@@ -43,6 +52,6 @@ def add_features( # pylint: disable=unused-argument
                 node.features[Nfeat.VARIANTRES] = residue.amino_acid.onehot
                 node.features[Nfeat.DIFFCHARGE] = 0
                 node.features[Nfeat.DIFFSIZE] = 0
-                node.features[Nfeat.DIFFPOLARITY] = numpy.zeros(residue.amino_acid.polarity.onehot.shape)
+                node.features[Nfeat.DIFFPOLARITY] = np.zeros(residue.amino_acid.polarity.onehot.shape)
```
cbaakman, line 55: Why does numpy have to be named np ?

cbaakman (commented):
Where did the code for parsing the forcefield parameters go?
Or have the forcefield parameter files been replaced with something else?

Nevermind! Found it!

deeprankcore/features/surfacearea.py
```diff
@@ -1,17 +1,57 @@
...
+        node.features[Nfeat.SASA] = area
+
+
 def add_features(pdb_path: str, graph: Graph, *args, **kwargs): # pylint: disable=too-many-locals, unused-argument
 
-    "calculates the buried surface area (BSA): the area of the protein, that only gets exposed in monomeric state"
+    """calculates the Buried Surface Area (BSA) and the Solvent Accessible Surface Area (SASA):
+    BSA: the area of the protein, that only gets exposed in monomeric state"""
 
+    # BSA
```
cbaakman, line 54: This section of code is quite long. Is there any possibility to make an `add_bsa` function?

deeprankcore/domain/edgefeatures.py
cbaakman: I see, they're not really feature names. They indicate what the hdf5 groups should be named. I think it's a bit confusing to have this module named  edgefeatures. Perhaps it can be `hdf5/storage/features.py` or similar.

gcroci2: I think it's fine that they are in the domain/ folder because they contain domain-related features (both for edges and nodes). Also, it's good to minimize the number of subtrees in the package and try to find a tradeoff between clearness, readability, and simplicity. What about `domain/edgestorage.py` and `domain/nodestorage.py`? @cbaakman @DaniBodor

deeprankcore/features/components.py
gcroci2: Very often it's imported as np, it's a simplification, and we changed it consistently (everywhere) in the codebase

deeprankcore/domain/edgefeatures.py
cbaakman: I think that would do fine.

gcroci2: Done, now we have edgestorage.py, nodestorage.py and targetstorage.py files containing the global variables used for naming Groups and Datasets in the hdf5 files

DaniBodor: i also like this: 

> What about `domain/edgestorage.py` and `domain/nodestorage.py`?

cbaakman (approved):

DaniBodor (approved):

PR #247, opened 2022-11-16
refactor: make `preprocess` use all available feature modules as default

gcroci2:
All available feature modules are used by `preprocess` function, thus generating all possible features in the hdf5 files. At a later stage, the user can select which ones to keep for training the networks.


cbaakman (commented):
Does this completely remove the option for the user to select a subset of features? Especially the atomic contact features are time-consuming to preprocess. So I can imagine that the user sometimes wants to skip them.

deeprankcore/preprocess.py
```diff
@@ -20,30 +22,35 @@ def _preprocess_one_query(prefix: str, feature_names: List[str], query: Query):
...
 
 
-def preprocess(feature_modules: List, queries: List[Query],
-               prefix: Optional[str] = None,
-               process_count: Optional[int] = None):
+def preprocess(
+    queries: List[Query],
+    prefix: Optional[str] = None,
+    process_count: Optional[int] = None,
+    feature_modules: Union[List[ModuleType], str] = "all"):
```
DaniBodor, line 37: I would maybe make feature_modules the second argument here. It's a core change to what happens in the function, whereas the other two arguments are less impactful.

DaniBodor:
> > Does this completely remove the option for the user to select a subset of features? Especially the atomic contact features are time-consuming to preprocess. So I can imagine that the user sometimes wants to skip them.
> 
> You're right, now I reinserted the possibility to select a subset in `preprocess`. The default is selecting all the modules in deeprankcore.features.

I guess it's still not possible to choose on a per feature level, only on a per module level.
But I think it's good enough like this, especially because each module uses only one (external) package, so probably not much savings on runtime by ignoring individual features within a module.

DaniBodor (approved):
Looks good!
I just had the one minor note above.

cbaakman (approved):

deeprankcore/preprocess.py
gcroci2: I'm merging the PR now, but we can discuss this further when you'll be back @DaniBodor

PR #250, opened 2022-11-18
refactor: save preprocessed data into one .hdf5 file as default

gcroci2:
Now the preprocessed data are saved into one hdf5 as default. The user can always decide to keep them in separated files using `combine_files` parameter in `preprocess` function. 

This is the current order of arguments in `preprocess`:
```
def preprocess(
    queries: List[Query],
    prefix: Optional[str] = None,
    process_count: Optional[int] = None,
    combine_files: bool = True,
    feature_modules: Union[List[ModuleType], str] = "all"
    ):
```

Now the user can fill in the arguments like this: `preprocess(queries, prefix)` in the default case (that is supposed to be the most used one, especially when starting using the package; the possibility to select only some features from the beginning should be used by "expert" users that already have lots of insights on the features themselves). So if we put it in a position different from the last ones --same applies to `combine_files`-- , then it will be always needed to specify the arguments like this: `preprocess(queries=queries, prefix=prefix)`.

But we can think about changing such order @DaniBodor


DaniBodor (approved):

deeprankcore/preprocess.py
```diff
@@ -34,7 +38,9 @@ def preprocess(
     queries: List[Query],
     prefix: Optional[str] = None,
     process_count: Optional[int] = None,
-    feature_modules: Union[List[ModuleType], str] = "all"):
+    combine_files: bool = True,
+    feature_modules: Union[List[ModuleType], str] = "all"
```
DaniBodor, line 42: why not use `None` as default make that use all features? Would look slightly cleaner that way, no? Or is it confusing that None is all?

gcroci2: No I think it's fine, if it's None then all features are used. Updated now. We should then do the same with other parameters (e.g., node_feature and edge_feature in DataSet: when None, all are used) @DaniBodor

tests/test_preprocess.py
```diff
@@ -51,22 +53,50 @@ def preprocess_tester(feature_modules: Union[List[ModuleType], str]):
...
+
+def test_preprocess_combine_files_true():
     """
+    Tests preprocessing for combining hdf5 files into one.
+    """
+
+    output_directory, output_paths = preprocess_tester(5, True, "all")
+    
+    assert len(output_paths) == 1
+
```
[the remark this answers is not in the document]
gcroci2: I inserted a check on the keys and values of the generated groups (see the new `test_preprocess_combine_files_true ` function)

deeprankcore/preprocess.py
DaniBodor: In those cases, I think it's theoretically possible to use no node or edge features and actually set it to None (e.g. if you only want to use node features but no edge features).
On second thought, maybe here that would also be true. Not sure if there would be any situation where it would make sense to run a Query with absolutely no features at all, though.

deeprankcore/query.py
```diff
@@ -117,6 +125,130 @@ def __repr__(self) -> str:
...
+
+        graph.write_to_hdf5(output_path)
+
+    def process( # pylint: disable=too-many-locals
+        self, 
+        prefix: Optional[str] = None,
+        process_count: Optional[int] = None,
+        combine_files: bool = True,
+        feature_modules: List[ModuleType] = None
+        ) -> List[str]:
```
DaniBodor, line 186: Regarding the order, I agree that queries and prefix should be first. But I think it's more likely for a user to want to set the modules than to change the process count or file combining.   
So maybe:
```
def preprocess(
    queries: List[Query],
    prefix: Optional[str] = None,
    feature_modules: Union[List[ModuleType], str] = "all"
    process_count: Optional[int] = None,
    combine_files: bool = True,
    ):
```

Also, maybe rename `combine_files` to `combine_output`

deeprankcore/query.py
```diff
@@ -117,6 +125,130 @@ def __repr__(self) -> str:
...
+                If None, all available modules in deeprankcore.features are used to generate the features. 
+                Defaults to None.
+        """
+
+        if process_count is None:
+            # returns the set of CPUs available considering the sched_setaffinity Linux system call,
+            # which limits which CPUs a process and its children can run on.
+            process_count = len(os.sched_getaffinity(0))
+
+        _log.info(f'\nSet of CPU processors available: {process_count}.')
```
DaniBodor, line 211: Should there be an error or warning raised if `process_count` is set higher than the number of available CPUs? I am assuming that that is not possible or at least not a good idea.

Or perhaps a just warning and then defaulted to number of CPUs anyway.

And I guess the log info should state CPUs _used_, not CPUs _available_

deeprankcore/preprocess.py
gcroci2: Let's keep None here, and the rest as it is. As soon as everything is documented, also different things are fine. I don't think there are any situations in which it would make sense to run a Query with no features at all, so None here makes sense to me.

deeprankcore/query.py
gcroci2: queries is no more a parameter (it's a property of the QueryCollection class now, I merged the other PR to this one). I modified the order of the former preprocess function, now process method, and the naming, like this:
```
    def process(
        self, 
        prefix: Optional[str] = None,
        feature_modules: List[ModuleType] = None,
        subprocesses: Optional[int] = None,
        combine_output: bool = True,
        ) -> List[str]:
```

deeprankcore/query.py
gcroci2: I added a warning that will be printed in the log file. Also, if the CPUs available are less than `subprocesses`, the latter is defaulted to the number of CPUs available.

deeprankcore/query.py
DaniBodor: I think `processes` as an argument to `process` is slightly confusing. That's why I suggested `subprocesses` above. If you don't like that, maybe `cpus` or `cpu_count` or something like that is better?

Issue #194, opened 2022-09-23
Ensure that `dataset_train` `dataset_val`, and `dataset_test` are all equivalent


DaniBodor:
I assume that for the training to work, all three HDF5DataSet objects have to have many (all?) the same properties (e.g. `edge_features`, `node_features`, `target`). 
I don't think there is currently any check for this (unless the train/val split is done internally, which will not be the typical way of doing experiments).

Ideally, these should only be set for `dataset_train` and then read by `dataset_val` and `dataset_test`

DaniBodor:
This probably implies moving a lot of stuff over from `HDF5DataSet` class to `Trainer` class.
This seems appropriate, given that `HDF5DataSet` is described as "Class from which the hdf5 datasets are loaded" and `Trainer` is described as "Class from which the network is trained, evaluated and tested". Most of these parameters are not directly related to the dataset itself, but rather to the network training.

Definitely move (including associated methods): `target`, `task`, `classes`, `node_features`, `edge_features`.
Unclear about:
- `root`, `transform`, `pre_transform`, `edge_features_transform`: see also issue #158
- `dict_filter`: I am unclear as to what this does
- `tqdm`: I need to check where this is implemented, but I guess on training rather than on loading
- `clustering_method`: I need to check when the clustering happens

DaniBodor:
Made `load_one_graph` use all available features/targets rather than just the selected ones.
However:
1. there is no test that allows me to check whether it works as intended --> create a unit test for this
2. I am not sure what the purpose of this function is, so I am unsure whether it's intended to load all possible features/targets or just the selected ones

DaniBodor:
after much trial and error, I realized that it is not possible to move the node and edge features to the trainer, because load_one_graph needs them and GraphDataset needs load_one_graph.
Instead, I'll raise an error if they do not match.

Issue #236, opened 2022-11-11
Restructure `HDF5DataSet` and `Trainer` classes


gcroci2:
### HDF5DataSet:
- Use a more generic name (Dataset?)
  - Settled on GraphDataset because pytorch-geometric already has a Dataset class
- Initialize with:
  - hdf5 path of raw data
  - keys of the hdf5 file that needs to be included in the specific HDF5DataSet instance
  - filtering options for features and/or targets
  - it should have `torch_geometric.data.data` Data object instance as an attribute

### Trainer
- Initialized with:
  - batch size, epochs, shuffle, ...
  - callback (early stopping)
  - metrics exporters
  - saving options
  - pretrained model
- Include a training method (called fit or train), that takes inputs
  - features
  - initialized neural network (Net)
  - HDF5Datasets 
  - targets/task

rename files to all lower case letters (as per python convention)

DaniBodor:
features, targets, etc need to be in Dataset class, due to the pay pytorch gemoetric handles these. Basically, the get function doesn't work as intended if the Dataset class contains features that are not used for training. Hence, the feature selection needs to be done on the Dataset class already.

Will make separate issue for moving attributes from Trainer class to Trainer.fit (or Trainer.train?) method: #234

Issue #152, opened 2022-08-05
Explore `target` parameter in HDF5DataSet


gcroci2:
In `HDF5DataSet` class, `target` parameter usage can be improved. Docs say that it can be `irmsd, lrmsd, fnat, bin, capri_class or DockQ. Defaults to None.`

- [ ] What happens if `target` is `None`? If it's not possible that there is no target (because there is no logic for it in `NeuralNet` class), we should throw an error in this case.
- [ ] What happens if the user sets a target not in the above list? If this is possible, it should be explicitly written in the docs. 
- [x] Now `task` parameter is set in `NeuralNet` class, according to `target` parameter, which is set in `HDF5DataSet` class. I think it's better to move `target` parameter to `NeuralNet` instead of `HDF5DataSet` class. Note also that conceptually, the target should be passed to either a class or function in a later stage, during the training process. This is another reason why I think it would be handy to move the `target` parameter to the `NeuralNet` class.
    - After discussion below: instead move `task` parameter to `HDF5DataSet` class (done in #200)
- [ ] After having decided about the above, and implemented that, remember to add an example code in the readme of `target` parameter usage.

cbaakman:
If target is None, then no ground truth is defined. It means that a model cannot be trained and only testing/predicting should work.
https://github.com/DeepRank/deeprank-core/blob/cb22435bad8c36024aef47cdf377b30fdb5a03ac/deeprankcore/DataSet.py#L334

If the user sets a target not in the above list, then theoretically it should still work, if that target is in the preprocessed data file. The user will have to manually specify a task as 'class' or 'reg' though.

The Dataset class needs the target name to find the ground truth values in the preprocessed data. That's most likely the reason why this variable must be provided to the class. Without it, the load_one_graph method wouldn't be able to produce any ground truth data.

DaniBodor:
Is this parameter all possible options of what is called `"score"` in storage.py?
If so, should it then also be renamed to `labels`, following the same logic as in issue #150?

DaniBodor:
> Now `task` parameter is set in `NeuralNet` class, according to `target` parameter, which is set in `HDF5DataSet` class. I think it's better to move `target` parameter to `NeuralNet` instead of `HDF5DataSet` class. Note also that conceptually, the target should be passed to either a class or function in a later stage, during the training process. This is another reason why I think it would be handy to move the `target` parameter to the `NeuralNet` class.

[...]

PR #253, opened 2022-11-23
refactor: simplify GraphDataset class

DaniBodor:
Make `GraphDataset` class more in line with `Dataset` class from torch_geometric.data.dataset

closes: #194
also partially solves #152 and #236


DaniBodor:
node- and edge-features cannot be moved out of dataset class. Made a new PR without that change instead of this one

Issue #237, opened 2022-11-11
Create basic transformations module


gcroci2:
It would be nice to have a module with some basic normalization/standardization functions that can be applied to different features of the HDF5Dataset instances.

Such operations should be performed after the definition of HDF5Dataset instances but before the Trainer fitting.

DaniBodor:
also reorganize a bit how these are handled by GraphDataset

gcroci2:
This issue may close #158

gcroci2:
Useful post https://towardsdatascience.com/the-best-format-to-save-pandas-data-414dca023e0d

PR #261, opened 2022-11-29
feat: add hist module for plotting raw hdf5 files features distributions

gcroci2:
Next step will be to implement the actual trasformations in the DeeprankDataset objects, see #237


deeprankcore/tools/transform.py
```diff
@@ -0,0 +1,199 @@
...
+
+    edge_features (list, optional): consider all pre-computed edge features ("all")
+        or some defined edge features (provide a list, example: ["dist", "coulomb"]).
+        The complete list can be found in deeprankcore/domain/edgestorage.py
+
+    target_features (list, optional): consider all pre-computed target features ("all")
+        or some defined target features (provide a list, example: ["binary", "capri_class"]).
+        The complete list (only of the pre-defined ones) can be found in deeprankcore/domain/targetstorage.py
+    Returns
+    ----------
```
DaniBodor, line 44:
```suggestion
        The complete list (only of the pre-defined ones) can be found in deeprankcore/domain/targetstorage.py

        Returns
    ----------
```

deeprankcore/tools/transform.py
```diff
@@ -0,0 +1,199 @@
...
+                    raise ValueError(
+                        f"The edge feature _{feat}_ was not found in the file {hdf5_path}.\
+                        \nAvailable edge features: {available_edge_features}"
+                    )
+            # check target features
+            for feat in target_features:
+                if feat not in available_target_features:
+                    raise ValueError(
+                        f"The target feature _{feat}_ was not found in the file {hdf5_path}.\
+                        \nAvailable target features: {available_target_features}"
```
DaniBodor, line 107: It seems to me that a large part of this is pretty much equivalent to the `_check_features` method of `GraphDataset`. Would it make sense to pull that function out of `GraphDataset` and make a generic one that can be called from different places to avoid redundancy?

Alternatively, depending on when in GraphDataset we call for transformation, it might not be necessary to do the checks again. Maybe we just make sure that transformation happens _after_ `_check_features`, and then it doesn't need to be checked again here.

deeprankcore/tools/transform.py
```diff
@@ -0,0 +1,199 @@
...
+
+    return df_final
+
+
+def save_hist(
+    df: pd.DataFrame,
+    features: Union[str,List],
+    fname: str,
+    bins: Union[int,List,str] = 10,
+    figsize: Tuple = (15, 15)
```
DaniBodor, line 145:
```suggestion
    df: pd.DataFrame,
    features: Union[str,List[str]],
    fname: str,
    bins: Union[int,List[float],str] = 10,
    figsize: Tuple = (15, 15)
```

DaniBodor: should we maybe default the bins to 'auto'?

DaniBodor (approved):
Really nice! This is really useful to be able to see the values/distributions of the features. Next step will be actually implementing the transformation. Maybe this module should be renamed, because (at least for now) it is not actually transforming anything, just visualizing; would data_visualization.py be a good name?

Regarding my comment about `_check_features`: don't know whether it is worthwhile dealing with/fixing this now, depending on how we will proceed with using hdf5 files or not,
Whichever way you decide to go on it is fine, but if you don't change it now, maybe open a new issue about it so we don't forget?

DaniBodor:
BTW, I haven't tested what the Dataframe and histogram outputs look like, but I assume you have and it looks as intended. Should I take a closer look at that as well?

gcroci2:
> Really nice! This is really useful to be able to see the values/distributions of the features. Next step will be actually implementing the transformation. Maybe this module should be renamed, because (at least for now) it is not actually transforming anything, just visualizing; would data_visualization.py be a good name?
> 
> Regarding my comment about `_check_features`: don't know whether it is worthwhile dealing with/fixing this now, depending on how we will proceed with using hdf5 files or not, Whichever way you decide to go on it is fine, but if you don't change it now, maybe open a new issue about it so we don't forget?

I renamed transform.py to hist.py (it converts the hdf5 files to pandas for plotting and saving histograms), and the same for the test script. When we'll have a clearer pipeline for the transformations we can improve the naming again.

deeprankcore/tools/transform.py
gcroci2: I inserted the line and I unified the comment to the [Google docstrings format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html), the one we are using in the modules already on readthedocs.

deeprankcore/tools/transform.py
gcroci2: The problem here is that the histograms visualization is done before creating the GraphDataset object, after the querying process. 

We have two options here: either we keep it this way, or we integrate this to_pandas and hist part into methods of the DeeprankDataset class. Actually I think the latter makes sense, and this way we're not going to have the _check_features repetition anymore. Also the parameters that hdf5_to_pandas takes as input are all covered by DeeprankDataset class. What do you think?

deeprankcore/tools/transform.py
gcroci2: Often auto doesn't work well, especially for a huge number of continuous data points. 10 is the default used by Matplotlib, so I would leave that :)

gcroci2:
> BTW, I haven't tested what the Dataframe and histogram outputs look like, but I assume you have and it looks as intended. Should I take a closer look at that as well?

I plotted the hists for all the features of the 140k datapoints and they look good, so don't worry :)

deeprankcore/tools/transform.py
DaniBodor: > Actually I think the latter makes sense

I agree,

> Anyway I would open an issue about this, because if we're not going to keep hdf5 files is not convenient to fix this now. If we'll keep the hdf5 format then it's just a matter of integrating the functions as methods in the above mentioned class.

Agree with this as well

DaniBodor:
> > Really nice! This is really useful to be able to see the values/distributions of the features. Next step will be actually implementing the transformation. Maybe this module should be renamed, because (at least for now) it is not actually transforming anything, just visualizing; would data_visualization.py be a good name?
> 
> I renamed transform.py to hist.py (it converts the hdf5 files to pandas for plotting and saving histograms), and the same for the test script. When we'll have a clearer pipeline for the transformations we can improve the naming again.

don't forget to rename the PR as well :)

DaniBodor:
I think for the one-hot encoded values, it would be more useful to plot a single histogram that shows how many of each is present, rather than separate 0-1 histograms for each option. 
This can also be a separate issue that we can tackle when we have time for it.

deeprankcore/tools/transform.py
gcroci2: Opened in #315

gcroci2:
> I think for the one-hot encoded values, it would be more useful to plot a single histogram that shows how many of each is present, rather than separate 0-1 histograms for each option. This can also be a separate issue that we can tackle when we have time for it.

Opened in #317 :)

PR #262, opened 2022-11-29
refactor: reorganize 

gcroci2:


deeprankcore/utils/metrics.py
```diff
@@ -27,11 +27,11 @@ def __exit__(self, exception_type, exception, traceback):
         pass # pylint: disable=unnecessary-pass
```
cbaakman, line 37: So your intention is to make the metrics exporter always take a directory path as input, but why isn't this an argument to the __init__ function?

gcroci2: > @gcroci2 I originally intended children of the `MetricsExporter` class to be added to a `MetricsExporterCollection` object within the `Trainer` object. This way, a metrics exporter can keep a file or a tensorboard writer open and close it at the end of the with-clause. You could add the `save_all_metrics` to the `__exit__` method, if you want it to be called at the end of a training/testing run.
> 
> I don't understand which if-statement you mean, sorry!

I updated the PR and my comment above, please have a look :)

gcroci2: Now it is :)

gcroci2:
This is still a draft, I'll ask again for your review when the PR is ready. For now, I just need to address my questions in the comment above (I've updated it) @cbaakman. Sorry for the confusion!

gcroci2:
> @gcroci2 I originally intended children of the `MetricsExporter` class to be added to a `MetricsExporterCollection` object within the `Trainer` object. This way, a metrics exporter can keep a file or a tensorboard writer open and close it at the end of the with-clause. 

So the enter and the exit methods are useful only considering the tensorboard exporter? I am still not getting why we need them for the other classes @cbaakman ; but yes sure I can use them to implement the save_all_outputs method and also the creation of the pandas dataframe (for ConciseOutputExporter).

This is still a draft, I'll ask again for your review when the PR is ready @cbaakman. Sorry for the confusion!

cbaakman:
Right now the tensorboard exporter is the only exporter using the enter and exit methods, but they can be placeholders for other actions that must happen at the beginning or end of a run. I think it's a good thing that the exporters get a call when the run starts/ends. For saving the output for example.

gcroci2:
**Summary of the PR**

_CI:_
- I updated pytorch and pytorch-geometric installation, both in the workflows and in the docs. For some reasons, pytorch-geometric was giving problems and the build was failing. Now we use pytorch 1.13 with CUDA 11.7, and everything works fine. We install both pytorch and pytorch-geometric through conda. 
- prospector started failing as well. I [...]
_General:_
- I renamed
  - metrics.py to exporters.py, test_metrics.py to test_exporters.py
  - `MetricsExporter` and `MetricsExporterCollection` to `OutputExporter` and `OutputExporterCollection`
  - the output directory in which exporters files are saved to output/ as default name
  - the output file created by `ConciseOutputExporter` to output_exporter.hdf5
  - `metrics_exporters` argument of `Trainer` class to `output_exporters`
  - `ConciseOutputExporter` to `HDF5OutputExporter`
- The output dir (output/) is now created in `OutputExporter` init (base class, parent of the other exporters)
- I made really optional the argument `target_data_shape` of `is_compatible_with`
- `output_exporters` argument of `Trainer` class defaults to `HDF5OutputExporter`
- `CSVOutputExporter` (former `OutputExporter`) has been deleted, together with its test
- docs have been updated accordingly
_HDF5OutputExporter:_
- I uniformed it to `OutputExporter` interface
  - `HDF5OutputExporter` makes use of `__enter__` and `__exit__` methods, making it a proper [context manager](https://peps.python.org/pep-0343/).
- Now results computed during training are saved in the hdf5 file in a Group called training, while results computed during the test are saved in a Group called testing. I am using now the append mode, which is better than overwriting the hdf5 file entirely (before I was updating the `df` attribute (pandas dataframe) of `HDF5OutputExporter` and overwriting the hdf5 file each time).
- `HDF5OutputExporter` has its own test in test_exporters.py

cbaakman (approved):

README.md
```diff
@@ -245,22 +247,19 @@ Let's define a Trainer instance, using for example of the already existing GNNs,
...
-metrics_exporters = [OutputExporter(metrics_output_directory)]
 
 trainer = Trainer(
     GINet,
     dataset_train,
     dataset_val,
     dataset_test,
-    batch_size = 64,
-    metrics_exporters = metrics_exporters
+    batch_size = 64
```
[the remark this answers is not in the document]
gcroci2: Depending on your data size and model architecture you may want to change batch_size (it can affect performances a lot)

Issue #265, opened 2022-12-06
Parallelize add Query operation


gcroci2:
Right now, to populate [QueryCollection](https://github.com/DeepRank/deeprank-core/blob/413d215a50370274bfc481b4a03c7ab937150827/deeprankcore/query.py#L134) we use the `add` method that basically only appends [Query](https://github.com/DeepRank/deeprank-core/blob/413d215a50370274bfc481b4a03c7ab937150827/deeprankcore/query.py#L32) instances to a list, `_queries` attribute. For few thousands of data this works pretty well, but now that we're starting having data in the order of hundreds of thousands this is not convenient anymore. We should parallelize such operations.

PR #266, opened 2022-12-06
perf: parallelize add Query to QueryCollection method

gcroci2:
See issue #265


gcroci2:
I realized that the bottleneck was not the list appending operation, but the matching between the pdb files and the pssms ones. Indeed I was calling `glob.glob` operation (to find the correspondent pssms files) for _each_ pdb file. Then for now there is no need to parallelize the add operation.

PR #288, opened 2022-12-14
refactor: fix `transform_sigmoid` logic and move it to `GraphDataset` class

gcroci2:
- Move `transform_sigmoid` parameter, now called `target_transform`, to GraphDataset class
- Only the target (ground truth) is now transformed with a log and a sigmoid function. The prediction will be done according to such a transformed target. 
- An error is raised if the user tries to set`target_transform` as True but the task is not regression
- Tests and docs have been added accordingly

### DIFF
--- a/deeprankcore/dataset.py
+++ b/deeprankcore/dataset.py
@@ -120,10 +120,9 @@ class GraphDataset(Dataset):
             edge_features_transform (function, optional): transformation applied to the edge features.
                 Defaults to lambdax:np.tanh(-x/2+2)+1.
 
-            target_transform (bool, optional): Bool to decide whether or not to apply a sigmoid transformation to the
-                target (for regression only). If True, first a log and then a sigmoid transformations
-                are applied to the target. This can result in a more uniform target distribution,
-                can speed up the optimization and puts target value between 0 and 1.
+            target_transform (bool, optional): Apply a log and then a sigmoid transformation to the target (for regression only).
+                This puts the target value between 0 and 1, and can result in 
+                a more uniform target distribution and speed up the optimization.
                 Defaults to False.
 
             target_filter (dictionary, optional): Dictionary of type [target: cond] to filter the molecules.
--- a/deeprankcore/dataset.py
+++ b/deeprankcore/dataset.py
@@ -9,7 +9,7 @@ from ast import literal_eval
 import torch
 from torch_geometric.data.dataset import Dataset
 from torch_geometric.data.data import Data
-from typing import Callable, List, Union
+from typing import Callable, List, Union, Optional
 from deeprankcore.domain import (edgestorage as Efeat, nodestorage as Nfeat,
                                 targetstorage as targets)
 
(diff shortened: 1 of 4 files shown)


cbaakman (approved):

deeprankcore/dataset.py
```diff
@@ -119,6 +120,12 @@ def __init__( # pylint: disable=too-many-arguments
             edge_features_transform (function, optional): transformation applied to the edge features.
                 Defaults to lambdax:np.tanh(-x/2+2)+1.
 
+            target_transform (bool, optional): Bool to decide whether or not to apply a sigmoid transformation to the
+                target (for regression only). If True, first a log and then a sigmoid transformations
+                are applied to the target. This can result in a more uniform target distribution,
+                can speed up the optimization and puts target value between 0 and 1.
+                Defaults to False.
+
```
DaniBodor, line 128: Maybe slightly clearer like this.
```suggestion
            target_transform (bool, optional): Apply a log and then a sigmoid transformation to the target (for regression only).
                This puts the target value between 0 and 1, and can result in 
                a more uniform target distribution and speed up the optimization.
                Defaults to False.
```

deeprankcore/dataset.py
```diff
@@ -234,12 +242,13 @@ def load_one_graph(self, fname, mol): # noqa
...
-                    except Exception as e:
-                        _log.error(e)
-                        _log.info('If your target variable contains categorical classes, \
-                        please convert them into class indices before defining the GraphDataset instance.')
+                    y = torch.tensor([grp[f"{targets.VALUES}/{self.target}"][()]], dtype=torch.float).contiguous().to(self.device)
+
+                    if self.task == targets.REGRESS and self.target_transform is True:
+                        y = torch.sigmoid(torch.log(y))
+                    elif self.task is not targets.REGRESS and self.target_transform is True:
+                        raise ValueError(f"Task is set to {self.task}. Please set it to regress to transform the target with a sigmoid.")
```
DaniBodor, line 250: Is it necessary to raise an error in this case? Maybe a warning/log input is sufficient and then `target_transform` is just ignored.

DaniBodor (commented):
much better!

Do you think this should be defaulted to True for regression tasks? It sounds like this transformation should be the standard.  
If so, the error needs to be handled differently for classification tasks (see also my comment on that). One option would be to default it to None in the Class, and if not explicitly changed by user, it re-defaults to True for regression tasks.

gcroci2:
> much better!
> 
> Do you think this should be defaulted to True for regression tasks? It sounds like this transformation should be the standard. If so, the error needs to be handled differently for classification tasks (see also my comment on that). One option would be to default it to None in the Class, and if not explicitly changed by user, it re-defaults to True for regression tasks.

I think it's better to leave False as default. Usually, it's better to not manipulate the target unless there is some specific issue with the data (like in our case the distribution was shifted towards high values, that's why they wanted to apply the log and then map to sigmoid)

deeprankcore/dataset.py
gcroci2: Done :)

deeprankcore/dataset.py
gcroci2: Not necessary, but conceptually is wrong. Also, I'll change this again with the transformations PR, which will include also target transformations. In case we can change this there :)

PR #292, opened 2022-12-15
Add the option to have a grid box of different x,y and z dimensions

cbaakman:


deeprankcore/utils/grid.py
```diff
@@ -27,26 +27,27 @@ class GridSettings:
...
+
+     - sizes: x, y, z sizes of the box in Å
+     - points_counts: the number of points on the x, y, z edges of the cube
+     - resolutions: the size in Å of one x, y, z edge subdivision. Also the distance between two points on the edge.
     """
 
-    def __init__(self, points_count: int, size: float):
-        self._points_count = points_count
-        self._size = size
+    def __init__(self, points_counts: np.array, sizes: np.array):
```
gcroci2, line 36: I think that it would be more user-friendly to pass-in a list instead of np.array. What if `point_counts` and `sizes` are both lists, then converted only internally to np.arrays? Same also for the @property methods, I would return a list instead of a np.array. What do you think?

gcroci2 (approved):
Great! Thank you. I have only one comment (see the review)

deeprankcore/utils/grid.py
cbaakman: Yes, lists would be better indeed.

PR #303, opened 2023-01-10
feat: add early stopping to Trainer.train

### LINKED ISSUE #216: Add Early stopping
Early stopping is a form of regularization used to avoid overfitting the training dataset. Early stopping keeps track of the validation loss; if the loss stops decreasing for several epochs in a row the training stops. We should have something like [this](https://github.com/Bjarten/early-stopping-pytorch).

DaniBodor:
This PR changes:
- add an adaptation of [this module](https://github.com/Bjarten/early-stopping-pytorch/blob/master/pytorchtools.py) for early stopping
  - I added gap functionality (check for difference between val loss and train loss)
  - other changes have to do with optional model saving, verbosity, status feedback messages, etc.
- change model saving and naming logic of `Trainer.train()` method
- improve docstrings & log info of `Trainer.train()` method
- add a unit test for earlystopping module
  - Not sure how useful this unit test is...

closes #216


DaniBodor:
The unit test I created only tests the basic logic of the early stopper using dummy loss values without training anything. 

Ideally we would also have a functional/integrated test to run on an actually trained model to check that losses are passed to and processed by `EarlyStopper` as intended? However, I am not sure how to implement this in a way that forces a training that will trigger the early stopping criteria in a controlled (and timely) manner.

gcroci2 (approved):

gcroci2:
> The unit test I created only tests the basic logic of the early stopper using dummy loss values without training anything.
> 
> Ideally we would also have a functional/integrated test to run on an actually trained model to check that losses are passed to and processed by `EarlyStopper` as intended? However, I am not sure how to implement this in a way that forces a training that will trigger the early stopping criteria in a controlled (and timely) manner. Any idea how to do this?

The only thing that comes to my mind is to set up an "high" patience number with a very small dataset (like 20, 30, 50, depending on how fast the training remains) and then assess that the model saved is at least after the patience epoch number.

deeprankcore/utils/earlystopping.py
```diff
@@ -0,0 +1,130 @@
...
+    def __init__( # pylint: disable=too-many-arguments
+        self,
+        patience: int = 10,
+        delta: Optional[float] = None,
+        maxgap: Optional[float] = None,
+        verbose: bool = True,
+        path: str = 'checkpoint.pt',
+        trace_func: Callable = print,
+    ):
+        """Terminate training if validation loss doesn't improve after a given patience or if a maximum gap between validation and training loss is reached.
```
DaniBodor, line 16:
```suggestion
        """
        Terminate training if validation loss doesn't improve after a given patience or if a maximum gap between validation and training loss is reached.
```

gcroci2 (approved):

deeprankcore/utils/earlystopping.py
```diff
@@ -1,64 +1,53 @@
...
 
         Args:
             patience (int): How long to wait after last time validation loss improved.
-                            Default: 10
+                Default: 10
             
-            delta (float, optional): Minimum change in the monitored quantity to qualify as an improvement.
-                            Default: None
+            delta (float): Minimum change required to reset the early stopping counter.
+                Default: None
```
gcroci2, line 21:
```suggestion
                Default: 0
```

gcroci2 (approved):

PR #312, opened 2023-01-12
test: make sure that the grid orientation is as in the original deeprank for `ProteinProteinInterfaceAtomicQuery`

cbaakman:
ProteinProteinInterfaceAtomicQuery has been altered to generate this output. It will use the pdb2sql `get_contact_atoms` function to select exactly the same set of atoms. Additionally, the grid has been centered at the mean position of this set of atoms.

This also requires the `_get_atomic_density_koes` function to be added to grid.py. This function will generate the atomic density as in the original deeprank.

A test has been added named `test_grid_orientation`. This will read the data from an old preprocessed hdf5 file and compares the contents with the output from the grid. Not all of it though, just the grid points and the mapping of the carbon-based atomic density feature.


deeprankcore/utils/buildgraph.py
```diff
@@ -36,24 +37,74 @@ def _add_atom_to_residue(atom, residue):
...
+
+def _add_atom_data_to_structure(structure: PDBStructure,  # pylint: disable=too-many-arguments, too-many-locals
+                                x: float, y: float, z: float,
+                                atom_name: str,
+                                altloc: str, occupancy: float,
+                                element_name: str,
+                                chain_id: str,
+                                residue_number: int,
+                                residue_name: str,
+                                insertion_code: str):
```
DaniBodor, line 56: Can you add a docstring for this.
Also, is there a reason that altloc and occupancy are on the same line?

DaniBodor (approved):
Looks OK to me, but some of the stuff is a bit over my head, so probably good if @gcroci2 takes a look as well.

gcroci2 (commented):
Great work! I also didn't get every single line of code but everything makes sense. The only concern I have is about ProteinProteinInterfaceResidueQuery, I think it's good to test the grid orientation for it as well right?

cbaakman:
> The only concern I have is about ProteinProteinInterfaceResidueQuery, I think it's good to test the grid orientation for it as well right?

I will look for a way to compare a deeprank-core residue-based grid to the original deeprank data. It's probably PSSM features that I need to compare in this case.

gcroci2:
> > The only concern I have is about ProteinProteinInterfaceResidueQuery, I think it's good to test the grid orientation for it as well right?
> 
> I will look for a way to compare a deeprank-core residue-based grid to the original deeprank data. It's probably PSSM features that I need to compare in this case.

From what I know, so far we're experimenting only with residue based graphs and grids, so I would say that this is the most important verification to do. I'll open another issue for this.

gcroci2 (approved):

PR #313, opened 2023-01-12
feat: allow for different loss functions other than the default

DaniBodor:
Option to use different loss functions is added.

For regression, all default loss functions in pytorch are working.

For classification only 2 loss functions work (CrossEntropyLoss and NLLLoss). This likely has something to do with formatting of `pred` and `target` that is non-uniform across loss functions. I raised a new issue (#318) to solve this, as it will require some careful consideration.

Also: I added 'BA' as a default target type for regression tasks, and fixed a mistake in test_dataset.py


deeprankcore/trainer.py
```diff
@@ -443,18 +441,65 @@ def configure_optimizers(self, optimizer = None, lr: float = 0.001, weight_decay
...
+                elif lossfunction not in losses.regression_losses:
+                    _invalid_loss()
+            self.lossfunction = lossfunction()
 
         elif self.task == targets.CLASSIF:
             # Assign weights to each class
-            self.weights = None
             if self.class_weights:
                 targets_all = []
                 for batch in self.train_loader:
```
DaniBodor, line 505: @gcroci2 , I found an error that is due to PR #311 but wasn't detected there. It only shows up if `class_weights` is set to True, which apparently wasn't the case in any previous test.

Basically, if `class_weights` is True, then it needs to have a `train.loader` defined. However, now the `train_loader` is only defined during training, which happens after the loss function is set.

Any suggestion how to resolve this without reverting to defining `train_loader` in init after all?

gcroci2: I've run the tests locally but I got many errors, even in tests that do not user class_weights parameter. Did you notice that as well? Solved those tests, I have an idea about how to move stuff without reinserting the loaders in the init.

gcroci2 (changes requested):
Tests fixing, then I can fix the one about the weights

deeprankcore/trainer.py
```diff
@@ -742,6 +804,19 @@ def _format_output(self, pred, target=None):
...
                 [self.classes_to_index[x] if isinstance(x, str) else self.classes_to_index[int(x)] for x in target]
             ).to(self.device)
+            
+            if isinstance(self.lossfunction, (nn.BCELoss, nn.BCEWithLogitsLoss)):
+                # pred must be in (0,1) range and target must be float with same shape as pred
+                pred = F.softmax(pred)
+                target = torch.tensor(
+                    [[0,1] if x == [1] else [1,0] for x in target]
+                ).float().to(self.device)
+                # raise ValueError('BCELoss and BCEWithLogitsLoss are currently not supported.\n\t')
```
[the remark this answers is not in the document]
gcroci2: If you're not 100% sure I would just raise an exception for now.

DaniBodor:
> I've run the tests locally but I got many errors, even in tests that do not user class_weights parameter. Did you notice that as well? Solved those tests, I have an idea about how to move stuff without reinserting the loaders in the init.

Yeah, I think I added another bug last evening when I was working late/tired. Those should be fixed now and locally I only have the one with weights is failing now.

deeprankcore/trainer.py
DaniBodor: > I've run the tests locally but I got many errors, even in tests that do not user class_weights parameter. Did you notice that as well? Solved those tests, I have an idea about how to move stuff without reinserting the loaders in the init.

Yeah, I think I added another bug last evening when I was working late/tired. Those should be fixed now and locally I only have the one with weights failing. Also the build is now only failing on the one test.

DaniBodor:
> Tests fixing, then I can fix the one about the weights

done.

deeprankcore/trainer.py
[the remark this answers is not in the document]
DaniBodor: @cbaakman, do you have any feedback on this?

gcroci2 (approved):

tests/test_set_lossfunction.py
```diff
@@ -0,0 +1,184 @@
...
+            dataset_test=trainer.dataset_train,
+            pretrained_model=model_path)
+        
+        return trainer_pretrained
+
+
+class TestTrainer(unittest.TestCase):
+    @classmethod
+    def setUpClass(class_):
+        class_.work_directory = tempfile.mkdtemp()
```
cbaakman, line 40: What is this directory used for?

gcroci2: I think it's not used here, so it can be deleted. Also I would call the class differently, referring to the loss function rather than to the Trainer only.

DaniBodor: yeah, I just copy pasted from the test_trainer and didn't pay close attention to unused stuff.

deeprankcore/trainer.py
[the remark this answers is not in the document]
cbaakman: How much freedom do we want to give to the user? If the user can define the network by himself, then I think it's his own responsibility to make sure the output is in 0-1 range. If the user is not allowed to make his own custom network, then softmax should automatically be added.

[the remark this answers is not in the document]
DaniBodor: OK, for now I will just not allow BCELoss. I already opened a new issue for dealing with more loss functions and will refer to this discussion there.

PR #319, opened 2023-01-16
fix: `num_workers` parameter of Dataloader object

gcroci2:
This PR:
- Fixes `num_workers` parameter usage of Dataloader, which allows to load data in parallel on multiple CPUs. Data are initially loaded to CPUs only, to allow such parallelization and to let the user use `num_workers`.
- Improves the handling of the cuda device, transferring data to GPU only during training (tested and working on Snellius).
- Makes facultative doing the checks on the hdf5 files that can take long times in case of bigger datasets, and do not need to be repeated in each experiment.
- Defaults the clustering method to None (again, don't know why it was defaulted to mcl again).

### DIFF
--- a/deeprankcore/dataset.py
+++ b/deeprankcore/dataset.py
@@ -31,7 +31,8 @@ class DeeprankDataset(Dataset):
                  root_directory_path: str,
                  transform: Union[Callable, None],
                  pre_transform: Union[Callable, None],
-                 target_filter: Union[Dict[str, str], None]
+                 target_filter: Union[Dict[str, str], None],
+                 check_integrity: bool
     ):
         """
         Parent class of :class:`GridDataset` and :class:`GraphDataset` which inherits from :class:`torch_geometric.data.dataset.Dataset`.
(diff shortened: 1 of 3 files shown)


DaniBodor:
Do you want to fix the build error before review or should I take a look already?

gcroci2:
> Do you want to fix the build error before review or should I take a look already?

Better first to fix the test, I'll assign you as reviewers again later

DaniBodor (approved):
Hopefully this'll make it a lot faster!

DaniBodor:
Is there any reason (or any problem with) sending target to device twice in a row?
It happens in line 746 and again in line 752. (I can't comment inline, because it was already there and isn't newly added or modified code).

cbaakman:
I see the device setting is removed from the dataset class. Where is it set now? Was this tested on a gpu?

gcroci2:
> Is there any reason (or any problem with) sending target to device twice in a row? It happens in line 746 and again in line 752. (I can't comment inline, because it was already there and isn't newly added or modified code).

It shouldn't be a problem but good catch, I removed the redundancy :)

gcroci2:
> I see the device setting is removed from the dataset class. Where is it set now? Was this tested on a gpu?

Indeed now the datasets and the dataloaders are loaded to the cpu (or CPUs in the case in which num_workers is higher than 0). This allows the dataloaders parallelization on CPUs.

Then, in the training loop, the batches are sent to the GPU and by setting pin_memory parameter to True such migration should be very fast.

Finally, at the end of each epoch iteration, results are brought back to the cpu for saving them.

I copied this logic from the old deeprank, and I've searched online (even if there in no clear guide about the best practices to handle cpus AND gpus in a pytorch pipeline). I think it makes sense though. I tested it on Snellius and it's working, also using gpu partition and different number of workers. It detects the cuda device and everything. 

@cbaakman

cbaakman (approved):

PR #323, opened 2023-01-18
feat: center the grids as in the old deeprank

cbaakman:
Grids must be given a center position on creation.

With this modification, the query object assigns a center position to the graph objects. For single residue variant graphs, this is the variant residue position. For protein-protein interaction graphs, this is the mean of the contacting atom positions.

When the graph is converted into a grid, this center position is passed on and it will determine the position of the grid points.

A test has been added to check that the center position is the same as in the original deeprank.

### DIFF
--- a/deeprankcore/utils/graph.py
+++ b/deeprankcore/utils/graph.py
@@ -3,7 +3,7 @@ import logging
 import numpy as np
 import h5py
 from deeprankcore.molstruct.atom import Atom
-from deeprankcore.molstruct.residue import Residue
+from deeprankcore.molstruct.residue import Residue, get_residue_center
 from deeprankcore.molstruct.pair import Contact, AtomicContact, ResidueContact
 from deeprankcore.utils.grid import MapMethod, Grid, GridSettings
 from deeprankcore.domain import (edgestorage as Efeat, nodestorage as Nfeat, 
@@ -98,6 +98,9 @@ class Graph:
 
         # targets are optional and may be set later
         self.targets = {}
+
+        # the center only needs to be set when this graph should be mapped to a grid.
+        self.center = np.array((0.0, 0.0, 0.0))
 
     def add_node(self, node: Node):
         self._nodes[node.id] = node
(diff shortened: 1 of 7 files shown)


DaniBodor (changes requested):
Can you fix the checks before review :)

cbaakman:
Checks have passed

deeprankcore/utils/graph.py
```diff
@@ -99,6 +99,9 @@ def __init__(self, id_: str):
         # targets are optional and may be set later
         self.targets = {}
 
+        # the center only needs to be set when this graph should be mapped to a grid.
+        self.center = np.array((0.0, 0.0, 0.0))
+
```
DaniBodor, line 104: Then why does the Graph class need a center attribute?

DaniBodor (approved):
Looks good to me. Just one comment about something that's unclear to me, but doesn't really matter.

gcroci2 (approved):
Seems good to me! Only a comment about inputs for GridSettings, I remember we agreed on using lists that then should be converted to numpy arrays internally.

deeprankcore/utils/graph.py
cbaakman: That's right! They will be overwritten once the center is known. The problem is that this center position can only be determined from the atoms in the structure. The method `write_as_grid_to_hdf5` only takes a graph as input, not a structure. So I chose to store the center in the graph, just like the node positions.

I agree that this is not a clean solution. Should we redesign the class-diagram?

cbaakman:
> Seems good to me! Only a comment about inputs for GridSettings, I remember we agreed on using lists that then should be converted to numpy arrays internally.

I can change this, no problem.

> Also, to be sure, the centering will work for both ppi residue level and ppi atomic level, right?

Yes, both query object types have the centering implemented. They both center on the mean position of the contact atoms.

tests/utils/test_graph.py
```diff
@@ -61,7 +61,7 @@ def test_graph_build_and_export(): # pylint: disable=too-many-locals
         graph.write_to_hdf5(hdf5_path)
 
         # export grid to hdf5
-        grid_settings = GridSettings(np.array((0, 0, 0)), np.array((20, 21, 21)), np.array((20.0, 21.0, 21.0)))
+        grid_settings = GridSettings(np.array((20, 21, 21)), np.array((20.0, 21.0, 21.0)))
```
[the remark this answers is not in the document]
cbaakman: I will make it lists.

DaniBodor (approved):

PR #324, opened 2023-01-18
refactor: delete all outputs from unit tests after run

DaniBodor:
The models created in some of the unit tests were stored inside the repo in a few different places.

Also, the assertion to check for output_exporter in test_trainer was recognizing output from previous unit tests. These are now cleared at the beginning of each test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,7 +20,7 @@ dist/*
 *.lzone
 *.ref_pairs
 
-# hdf5
+# hdf5 and tar
 *.hdf5
 *.tar
 
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -15,7 +15,6 @@ from deeprankcore.tools.target import compute_targets
 from deeprankcore.domain import (edgestorage as Efeat, nodestorage as Nfeat,
                                 targetstorage as targets)
 
-model_path = './tests/test.pth.tar'
 
 def test_integration_cnn(): # pylint: disable=too-many-locals
     """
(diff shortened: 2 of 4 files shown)


gcroci2 (approved):
Great that now we delete files between the tests! Only a doubt: for tests like `test_fout`, `test_sgat`, ... we didn't assert the number of files in the working dir, but we do save the models. Shouldn't this be a problem for successive tests like `test_cuda`, which asserts the number of files in the dir? I guess we don't catch it because we're not really testing cuda. If this makes sense, I would suggest to either remove the files in each test generates one or just to move on top all the tests that assert the number of files in the temporary dir.

DaniBodor:
> Great that now we delete files between the tests! Only a doubt: for tests like `test_fout`, `test_sgat`, ... we didn't assert the number of files in the working dir, but we do save the models. Shouldn't this be a problem for successive tests like `test_cuda`, which asserts the number of files in the dir? I guess we don't catch it because we're not really testing cuda. If this makes sense, I would suggest to either remove the files in each test generates one or just to move on top all the tests that assert the number of files in the temporary dir.

In each of the tests that asserts the number of files after completion, i added an assert folder empty at the start of the test, so it should be covered. Nonetheless, you're probably right that it would make sense to make each test follow the same logic. I just adjusted the ones where this was already the case, but not the ones that didn't have this. I'll add this.
And indeed, specifically the cuda test does not currently run, given that the first if statement is false.

gcroci2 (approved):

PR #327, opened 2023-01-24
ci: do not close stale issues or PRs

DaniBodor:
I think we actually want old issues to just stay open, right? Until now, we were constantly removing the stale label. 

I think better to just leave as stale forever, so it's still clear that this is an "old" issue, but not close it until it's done. This way we can occasionally check the stale issues and see whether they have been done in the meantime or whether it is time to pick one up.

### DIFF
--- a/.github/workflows/stale_issue_pr.yml
+++ b/.github/workflows/stale_issue_pr.yml
@@ -13,12 +13,12 @@ jobs:
       - uses: actions/stale@v5.0.0
         with:
           days-before-issue-stale: 30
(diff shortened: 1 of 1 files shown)


DaniBodor:
what would be a good [conventional commit type ](https://www.conventionalcommits.org/en/v1.0.0/)for this?

gcroci2 (approved):
Totally agreed

gcroci2:
> conventional commit type

`ci:` fits I think.

Issue #331, opened 2023-01-27
Add a feature module for residue contact density


DaniBodor:
[Residue contact density (RCD) is a feature of the CNN version of deeprank](https://github.com/DeepRank/deeprank/blob/master/deeprank/features/ResidueDensity.py), and should also be maintained as a feature in deeprankcore. As far as I can tell, it is basically the number of residues on the opposite chain within a given cutoff radius.

The original deeprank has separate RCD features for 'total' and each of 6 combinations for 'apolar', 'polar', and 'charged'.  
I think these combinatorial ones can be brought back to 3, because each residue itself is of one type and so only has 3 options to combine with the other types.  
On the other hand, no distinction is made in deeprank (CNN) between positively and negatively charged residues, it is just called 'charged' (at first glance, this seems to be true throughout), but in deeprankcore we separate these into positive and negative charge. I guess this distinction between pos and neg should be maintained in the RCD features.

See also discussion in [issue 279](https://github.com/DeepRank/deeprank-core/issues/279#issuecomment-1404737582).

DaniBodor:
Few points of note:
- What is a good distance cutoff here? 
  - In deeprank a cutoff of 5.5 Å is used.
  - In theory separate features could be created for different distance cutoffs residue depth.
  - Do we want to make this a hard-coded parameter or a variable that can be set by users?
- Surrounding residues need to be read from the pdb files rather than the hdf5 files, because otherwise residues that are not part of the "interface" will not be counted.

PR #333, opened 2023-01-27
feat: add inter-residue contact (IRC) node features

DaniBodor:
Add inter-residue contact feature:
- residue contact density is number of contact residues _on the opposite chain_ within a cutoff distance, i.e. the contact residues
- separate features for total contacts as well as contacts of each polarity type
- see #331 for details

Also:
- found fixed error in `process` method of `Querycollection`.
- made style of `add_features` functions uniform among feature modules

TO DO:
- [x] calculate residue contact numbers
- [x] implement calculated "densities" as features
- [x] automatically detect chain ids
- [x] create test for this feature:
  - [x] nan values, 
  - [x] IRC_total == sum of separate IRCs
  - [x] non-zero total contacts
  - [ ] anything else? (actual values?)
- [x] consider [these issues](https://github.com/DeepRank/deeprank-core/issues/331#issuecomment-1406472678)
  - [x] How to define the cutoff value:
    1. :white_check_mark: hard coded --> 5.5A is based on results from [this paper](https://elifesciences.org/articles/07454)
    2. identical to the query cutoff 
    3. open the setting to the user
      - this would be nice in theory, but doesn't easily fit our `add_features` format
  - [x] should we raise error/warning for no/few total contacts, as was done in Deeprank? --> yeah, why not....
  - [x] this is not a true density, but an absolute number: can we rename the feature accordingly or is the naming a convention? --> renamed to inter-residue contacts, following [this paper](https://elifesciences.org/articles/07454)
  - [x] how to organize this feature:
    1. only total IRC
    2. total + polarity of partner
    3. :white_check_mark: total + polarity-pairing (as in original DeepRank) --> otherwise the IRC_Positive has a very different meaning for a Positive node (repulsive) than for a Negative node (attractive).
  - [x] do we need an analogous feature for atomic contacts as well? --> NO
    - if so, what does it group by?
- [x] decide what to do with equivalent `get_residue_contact_pairs` function in buildgraph.py (only used in some tests):
  - :white_check_mark: move code from irc.py to buildgraph.py, and call that from both tests and irc.py. I created a new issue for this #394 (not a priority!)
- ~reconsider nomenclature for vdw and electrostatic edge features (discussion on slack with Li & Coos)~
  - this will be dealt with in #357 instead


DaniBodor:
@gcroci2 , there seems to be a problem in the test_querycollection module, when adding PPI queries and specifically when `combine_output=False`. I am guessing it has to do with repeatedly adding the same file, which is a quirk of the way we are running the test.
As we were not previously testing PPI queries in this test (and/or because part of the test was wrapped in a try-block), we weren't catching it before.

Also something funky is happening with renaming the files (it goes '_2_2_2', instead of numbering up).

Any chance you can take a look at this.

gcroci2:
It took a while to understand what was going on. The numbering was incorrect (we were adding _2 instead of increasing the numbering as expected), but this was not the issue causing the error. The issue was to create only one instance for both residue and variant queries and then modify it in place instead of creating different copies of instances from the same object. The collection variable was containing queries pointing to the same query instance. I am modifying everything to fix both the numbering and the objects issues @DaniBodor

gcroci2:
test_querycollection fixed now :) @DaniBodor

DaniBodor:
> test_querycollection fixed now :) @DaniBodor

Thanks, I'll get back to work on the rest of this PR. Should be quick now

deeprankcore/features/rcd.py
```diff
@@ -97,15 +102,13 @@ def count_residue_contacts(pdb_path: str, chains: List[str], cutoff: float = 5.5
 def add_features(
     pdb_path: str, 
     graph: Graph,
-    *args, **kwargs): # pylint: disable=unused-argument
+    single_amino_acid_variant = None):
```
gcroci2, line 105: type hint missing

deeprankcore/query.py
```diff
@@ -236,15 +243,16 @@ def process( # pylint: disable=too-many-arguments, too-many-locals
                 Defaults to None, which means that all available modules in `deeprankcore.features` are used to generate
                 the features. 
             cpu_count (Optional[int], optional): How many processes to be run simultaneously. Defaults to None, which takes all available cpu cores.
-            combine_output (bool, optional): For combining the .HDF5 files generated by the processes. Defaults to True.
+            combine_output (bool, optional): For combining the .hdf5 files generated by the processes. Defaults to True.
```
gcroci2, line 246:
```suggestion
            combine_output (bool, optional): For combining the HDF5 files generated by the processes. Defaults to True.
```

deeprankcore/query.py
```diff
@@ -236,15 +243,16 @@ def process( # pylint: disable=too-many-arguments, too-many-locals
...
                 Defaults to None.
             grid_map_method (Optional[:class:`MapMethod`], optional): If valid together with `grid_settings`, the grid data will be stored as well.
                 Defaults to None.
-            grid_augmentation_count (int, optional): Number of grid data augmentations, must be zero or a positive number. Defaults to 0.
+            grid_augmentation_count (int, optional): Number of grid data augmentations. May not be negative be zero or a positive number.
+                Defaults to 0.
         
         Returns:
-            List[str]: The list of paths of the generated .HDF5 files.
+            List[str]: The list of paths of the generated .hdf5 files.
```
gcroci2, line 255:
```suggestion
            List[str]: The list of paths of the generated HDF5 files.
```

deeprankcore/tools/target.py
```diff
@@ -102,7 +102,7 @@ def compute_targets(pdb_path: str, reference_pdb_path: str) -> Dict[str, Union[f
     """
 
     ref_name = os.path.splitext(os.path.basename(reference_pdb_path))[0]
-    sim = StructureSimilarity(pdb_path, reference_pdb_path)
+    sim = StructureSimilarity(pdb_path, reference_pdb_path, enforce_residue_matching=False)
```
DaniBodor, line 105: newest version of pdb2sql raises an error if this is not suppressed. I have created a new issue (#395) to check whether this is something that needs to be dealt with rather than suppressed.

deeprankcore/features/irc.py
```diff
@@ -31,8 +32,9 @@ class _ContactDensity:
     """Internal class that holds contact density information for a given residue.
     """
     
-    def __init__(self, residue):
+    def __init__(self, residue, polarity):
```
gcroci2, line 35: May be useful to have type hints here as well even if it's an internal class

gcroci2 (approved):

PR #334, opened 2023-01-30
fix: gpu usage

gcroci2:
Fixes GPU runs which were failing due to tensors not being copied to the cpu for evaluation phase.


DaniBodor (approved):
Glad this was fixed (didn't realize there was a problem :P)

PR #336, opened 2023-02-02
feat: add data augmentation for grids

cbaakman:
Data augmentation rotates the features in the graph around a given axis, before mapping them to the grid.
A differently orientated version of that feature is added to the grid data entries in the hdf5 file.

A test was included to verify that the feature value remains the same, despite the different orientation.


deeprankcore/utils/grid.py
```diff
@@ -24,6 +27,22 @@ class MapMethod(Enum):
     NEAREST_NEIGHBOURS = 4
 
 
+class Augmentation:
+    "a rotation around an axis, to be applied to a feature before mapping it to a grid"
```
gcroci2, line 31:
```suggestion
    "A rotation around an axis, to be applied to a feature before mapping it to a grid."
```

deeprankcore/query.py
```diff
@@ -212,7 +220,8 @@ def process( # pylint: disable=too-many-arguments, too-many-locals
         cpu_count: Optional[int] = None,
         combine_output: Optional[bool] = True,
         grid_settings: Optional[GridSettings] = None,
-        grid_map_method: Optional[MapMethod] = None
+        grid_map_method: Optional[MapMethod] = None,
+        grid_augmentation_count: Optional[int] = 0
```
gcroci2, line 224: Optional[X] is equivalent to X | None (or Union[X, None]).

Note that this is not the same concept as an optional argument, which is one that has a default. An optional argument with a default does not require the `Optional` qualifier on its type annotation just because it is optional. 

So either we set the default to `None` (modifying the code accordingly) or we keep 0 removing the `Optional` hint.

gcroci2 (approved):
Very nice! I would just add a test to verify that in augmentation case we generate the expected number of data points (original number + number of data augmented). It would be handy to also verify that when the augmented hdf5 files are loaded to a GridDataset instance, the len in right there as well. Then feel free to merge if all the tests pass.

gcroci2 (approved):

PR #339, opened 2023-02-03
Check that the grid for residue-based protein-protein interfaces has the same center and orientation as in the original deeprank.

cbaakman:
The grid points are compared. They may be off by one Angstrom.


gcroci2 (approved):

gcroci2:
Great! Please fix the linting before merging @cbaakman

PR #353, opened 2023-02-14
docs: uniform docstrings format in query module

joyceljy:


deeprankcore/query.py
```diff
@@ -112,19 +111,15 @@ def model_id(self, value):
...
         return f"{type(self)}({self.get_query_id()})"
 
 
 class QueryCollection:
-    """
-    Represents the collection of data queries.
-        Queries can be saved as a dictionary to easily navigate through their data.
-    
-    """
+    """Represents the collection of data queries.Queries can be saved as a dictionary to easily navigate through their data."""
```
gcroci2, line 122:
```suggestion
    """Represents the collection of data queries. Queries can be saved as a dictionary to easily navigate through their data."""
```

deeprankcore/query.py
```diff
@@ -508,19 +473,16 @@ def __hash__(self) -> hash:
 
     @staticmethod
     def _get_atom_node_key(atom) -> str:
-        """Pickle has problems serializing the graph when the nodes are atoms,
-        so use this function to generate an unique key for the atom"""
+        """Pickle has problems serializing the graph when the nodes are atoms,so use this function to generate an unique key for the atom"""
```
gcroci2, line 476:
```suggestion
        """Since pickle has problems serializing the graph when the nodes are atoms, this function can be used to generate a unique key for the atom."""
```

gcroci2 (changes requested):
Great first PR :) Few comments:
- Please check that the hint in the method/function is the same in the doc string. For example, in line 481 `def build(self, feature_modules: List, include_hydrogens: bool = False)` features_modules should be as in the doc string `List[ModuleType]`, and not only `List`.
- Be sure to have checked usage of Optional type hinting. All the parameters that can have None (either as default or not default) should be hinted as Optional (see [typing.Optional](https://docs.python.org/3/library/typing.html#:~:text=On%20the%20other%20hand%2C%20if%20an%20explicit%20value%20of%20None%20is%20allowed%2C%20the%20use%20of%20Optional%20is%20appropriate%2C%20whether%20the%20argument%20is%20optional%20or%20not.%20For%20example%3A) docs). And make sure that the Optional already present can be really None.
- Maybe you've already checked, but be sure that the spacing in the doc strings for the type is the same (e.g. `param(type1, optional): [...]` instead of `param (type1, optional): [...]`). Both ways are fine, just pick one and be consistent across all PRs and files. 

Note that these comments apply also to all the other PRs, so please check them in this regard.

deeprankcore/query.py
```diff
@@ -219,21 +219,21 @@ def process( # pylint: disable=too-many-arguments, too-many-locals
...
+            feature_modules (Optional[List[ModuleType]], optional): List of features' modules used to generate features. Each feature's module must
                 implement the :py:func:`add_features` function, and features' modules can be found (or should be placed in case of a custom made feature)
                 in `deeprankcore.features` folder. Defaults to None, which means that all available modules in `deeprankcore.features` are used to generate
                 the features. 
-            cpu_count (int, optional): How many processes to be run simultaneously. Defaults to None, which takes all available cpu cores.
-            combine_output (bool, optional): For combining the .HDF5 files generated by the processes, defaults to True.
-            grid_settings (:class:`GridSettings`, optional): If valid together with `grid_map_method`, the grid data will be stored as well.
+            cpu_count (Optional[int], optional): How many processes to be run simultaneously. Defaults to None, which takes all available cpu cores.
+            combine_output (Optional[bool], optional): For combining the .HDF5 files generated by the processes. Defaults to True.
```
gcroci2, line 228:
```suggestion
            combine_output (bool, optional): For combining the .HDF5 files generated by the processes. Defaults to True.
```

gcroci2, line 228: Please change line 215 to `combine_output: bool = True,`
Indeed combine_output can't be actually None, but only True or False

deeprankcore/query.py
```diff
@@ -414,11 +414,11 @@ def __init__(  # pylint: disable=too-many-arguments
             insertion_code (str): The insertion code of the variant residue, set to None if not applicable.
             wildtype_amino_acid (:class:`AminoAcid`): The wildtype amino acid.
             variant_amino_acid (:class:`AminoAcid`): The variant amino acid.
-            pssm_paths (Dict(str,str), optional): The paths to the .PSSM files, per chain identifier. Defaults to None.
-            radius (float, optional): In Ångström, determines how many residues will be included in the graph. Defaults to 10.0. 
-            distance_cutoff (float, optional): Max distance in Ångström between a pair of atoms to consider them as an external edge in the graph.
+            pssm_paths (Optional[Dict(str,str)], optional): The paths to the .PSSM files, per chain identifier. Defaults to None.
+            radius (Optional[float], optional): In Ångström, determines how many residues will be included in the graph. Defaults to 10.0.
```
gcroci2, line 418: Are we sure that radius can be None? Just check it to be sure :)

gcroci2 (changes requested):

deeprankcore/query.py
joyceljy: I found that it cannot be None, I removed the Optional datat type.

gcroci2 (approved):

PR #354, opened 2023-02-14
docs: uniform docstrings format in trainer module

joyceljy:


gcroci2 (changes requested):
Great job! Be sure to:
- Have checked usage of `Optional` type hinting. All the parameters defaulted to None should be hinted as Optional (see [typing.Optional](https://docs.python.org/3/library/typing.html#:~:text=On%20the%20other%20hand%2C%20if%20an%20explicit%20value%20of%20None%20is%20allowed%2C%20the%20use%20of%20Optional%20is%20appropriate%2C%20whether%20the%20argument%20is%20optional%20or%20not.%20For%20example%3A) docs). And make sure that the Optional already present can be really None. 
- Please check that the hit in the method/function is the same in the doc string. For example, in query.py line 481 `def build(self, feature_modules: List, include_hydrogens: bool = False)` features_modules should be as in the doc string `List[ModuleType]`, and not only `List`.

deeprankcore/trainer.py
```diff
@@ -23,11 +23,11 @@ class Trainer():
     def __init__( # pylint: disable=too-many-arguments # noqa: MC0001
                 self,
                 neuralnet = None,
-                dataset_train: Union[GraphDataset, GridDataset] = None,
-                dataset_val: Union[GraphDataset, GridDataset] = None,
-                dataset_test: Union[GraphDataset, GridDataset] = None,
-                val_size: Union[float, int] = None,
-                test_size: Union[float, int] = None,
+                dataset_train: Union[GraphDataset, GridDataset,None] = None,
```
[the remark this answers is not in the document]
DaniBodor: It actually boils down to the same thing, but I think in most (or all) of our cases, the way Giulia suggested it is preferred

See the comment to [this post](https://stackoverflow.com/questions/69185886/preferred-python-typing-syntax-of-optionalunion-vs-unionnone), stating:

> I think it depends on the context, but I generally prefer Optional[...] in most cases where None is a default value for an argument that may or may not be provided a non-trivial value. But there are also cases where an argument or variable might not be "optional" where None is a valid value it can take. 

I noticed that python 10 has a [nicer way of dealing with Union](https://peps.python.org/pep-0604/), but I am currently running python 3.9 and I think that we don't want to force 3.10 on users, right?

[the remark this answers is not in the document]
gcroci2: Exactly I've already seen it and it is nicer, but since we're still supporting python 3.9 we must keep using Optional.

deeprankcore/query.py
```diff
@@ -283,8 +283,7 @@ def process( # pylint: disable=too-many-arguments, too-many-locals
         if combine_output:
             for output_path in output_paths:
                 with h5py.File(f"{prefix}.hdf5",'a') as f_dest, h5py.File(output_path,'r') as f_src:
-                    for key, value in f_src.items():
-                        _log.debug(f"copy {key} from {output_path} to {prefix}.hdf5")
+                    for _, value in f_src.items():
```
```suggestion
                    for key, value in f_src.items():
                        _log.debug(f"copy {key} from {output_path} to {prefix}.hdf5")
```

gcroci2 (changes requested):
Please commit my suggested change, then I'll give a very fast look again and then we can merge :)

deeprankcore/trainer.py
```diff
@@ -443,18 +425,18 @@ def configure_optimizers(self, optimizer = None, lr: float = 0.001, weight_decay
 
     def set_lossfunction(self, lossfunction = None, override_invalid: bool = False): #pylint: disable=too-many-locals # noqa: MC0001
```
gcroci2, line 426:
```suggestion
    def set_lossfunction(self, lossfunction: Optional = None, override_invalid: bool = False): #pylint: disable=too-many-locals # noqa: MC0001
```

deeprankcore/trainer.py
```diff
@@ -524,31 +506,30 @@ def train( # pylint: disable=too-many-arguments, too-many-branches, too-many-loc
...
         Args:
-            nepoch (int): Maximum number of epochs to run.
-                        Default: 1.
+            nepoch (int, optional): Maximum number of epochs to run.
+                        Defaults to 1.
             batch_size (int, optional): Sets the size of the batch.
-                        Default: 32.
+                        Defaults to 32.
             shuffle (bool, optional): Whether to shuffle the training dataloaders data (train set and validation set).
                         Default: True.
```
gcroci2, line 517:
```suggestion
                        Defaults to True.
```

gcroci2 (changes requested):

PR #355, opened 2023-02-14
docs: uniform docstrings format in dataset module

joyceljy:


deeprankcore/dataset.py
```diff
@@ -405,30 +388,23 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
...
             target (str, optional): Default options are irmsd, lrmsd, fnat, binary, capri_class, dockq, and BA. It can also be a custom-defined target
                 given to the Query class as input (see: `deeprankcore.query`); in this case, the task parameter needs to be explicitly specified as well.
                 Only numerical target variables are supported, not categorical. If the latter is your case, please convert the categorical classes into
                 numerical class indices before defining the :class:`GraphDataset` instance. Defaults to None.
-
             task (str, optional): 'regress' for regression or 'classif' for classification. Required if target not in
                 ['irmsd', 'lrmsd', 'fnat', 'binary', 'capri_class', 'dockq', or 'BA'], otherwise this setting is ignored.
                 Automatically set to 'classif' if the target is 'binary' or 'capri_classes'.
                 Automatically set to 'regress' if the target is 'irmsd', 'lrmsd', 'fnat', 'dockq' or 'BA'.
-
```
gcroci2, line 425: In the line below a square bracket is missing: node_features (Union[List[str], str`]`, optional). Same for edge_features below

gcroci2 (changes requested):
Great :)
Few comments:
- as in PR #354, be sure that the spacing in the doc strings for the type is the same (e.g. param(type1, optional): [...] instead of param (type1, optional): [...]). Both ways are fine, just pick one and use it for all files in all PRs. This comment also applies to the other PRs.
- The PR states that it's about dataset module, but then I see edits also in query and trained modules. Maybe you branched from your previous branch and that why I see these changes here as well? As a best practice, you should keep isolated the PRs. Meaning that, for example, first you branch from main to modify query and trainer, then you open a PR about it. Then you go again to main, and you branch again from there for opening a PR regarding another file, e.g. dataset. In this way in each PR only changes relative to it can be seen, and the rest will be merged separately. If this is not clear is super fine, we can have a call :)

deeprankcore/query.py
```diff
@@ -283,8 +283,7 @@ def process( # pylint: disable=too-many-arguments, too-many-locals
         if combine_output:
             for output_path in output_paths:
                 with h5py.File(f"{prefix}.hdf5",'a') as f_dest, h5py.File(output_path,'r') as f_src:
-                    for key, value in f_src.items():
-                        _log.debug(f"copy {key} from {output_path} to {prefix}.hdf5")
+                    for _, value in f_src.items():
```
```suggestion
                    for key, value in f_src.items():
                        _log.debug(f"copy {key} from {output_path} to {prefix}.hdf5")
```

gcroci2 (changes requested):
Great :) please commit my proposed changes and then I'll have a fast check again, then we can merge :)

gcroci2 (approved):

PR #356, opened 2023-02-14
docs: uniform docstrings format in features subpackage

joyceljy:


gcroci2 (changes requested):
Great! Similar comments:
- as in PR #354, be sure that the spacing in the doc strings for the type is the same (e.g. param(type1, optional): [...] instead of param (type1, optional): [...]). Both ways are fine, just pick one and use it for all files in all PRs. This comment also applies to the other PRs.
- Also here I see edits also in query, trainer and dataset modules. As a best practice, you should keep isolated the PRs (see my comment in PR #355). Meaning that, for example, first you branch from main to modify query and trainer, then you open a PR about it. Then you go again to main, and you branch again from there for opening a PR regarding another file, e.g. dataset. To solve this in the current PRs, you can just copy paste the files not included in the PR in your current branch (in this specific case: query, trainer, dataset), commit and push. Then in the Files changed tab of the PR on Github you'll see only files in features folder changed, which is exactly what we're looking for.

deeprankcore/dataset.py
```diff
@@ -372,7 +372,7 @@ def load_one_grid(self, hdf5_path: str, entry_name: str) -> Data:
 
             mapped_features_group = entry_group[gridstorage.MAPPED_FEATURES]
             for feature_name in self.features:
-                feature_data.append(mapped_features_group[feature_name][:])
+                feature_data.append(mapped_features_group[feature_name][gridstorage.FEATURE_VALUE][:])
```
gcroci2, line 375:
```suggestion
                feature_data.append(mapped_features_group[feature_name][:])
```

deeprankcore/query.py
```diff
@@ -283,8 +283,7 @@ def process( # pylint: disable=too-many-arguments, too-many-locals
         if combine_output:
             for output_path in output_paths:
                 with h5py.File(f"{prefix}.hdf5",'a') as f_dest, h5py.File(output_path,'r') as f_src:
-                    for key, value in f_src.items():
-                        _log.debug(f"copy {key} from {output_path} to {prefix}.hdf5")
+                    for _, value in f_src.items():
```
```suggestion
                    for key, value in f_src.items():
                        _log.debug(f"copy {key} from {output_path} to {prefix}.hdf5")
```

gcroci2 (changes requested):
Also here, please commit my suggested changes, then I'll give a fast look, then we can merge :)

gcroci2 (approved):

PR #359, opened 2023-02-15
docs: uniform docstrings format in neuralnets subpackage

joyceljy:


gcroci2 (changes requested):
Please remember to check out the general comments in PR https://github.com/DeepRank/deeprank-core/pull/353 and PR https://github.com/DeepRank/deeprank-core/pull/354, which apply to all PRs.

deeprankcore/neuralnets/gnn/foutnet.py
```diff
@@ -13,16 +13,16 @@ class FoutLayer(torch.nn.Module):
...
     This layer is described by eq. (1) of
     Protein Interface Predition using Graph Convolutional Network
-    by Alex Fout et al. NIPS 2018
+    by Alex Fout et al. NIPS 2018.
 
     Args:
         in_channels (int): Size of each input sample.
         out_channels (int): Size of each output sample.
-        bias (bool, optional): If set to :obj:`False`, the layer will not learn
+        bias (bool): If set to :obj:`False`, the layer will not learn
```
gcroci2, line 21:
```suggestion
        bias (bool, optional): If set to :obj:`False`, the layer will not learn
```

deeprankcore/neuralnets/gnn/foutnet.py
```diff
@@ -13,16 +13,16 @@ class FoutLayer(torch.nn.Module):
...
             an additive bias. (default: :obj:`True`)
```
gcroci2, line 22:
```suggestion
            an additive bias. Default: :obj:`True`.
```

deeprankcore/neuralnets/gnn/sgat.py
```diff
@@ -17,14 +17,15 @@ class SGraphAttentionLayer(torch.nn.Module):
     Ni is the number of neighbor of node i
     \\Sum_j runs over the neighbors of node i
     a_ij is the edge attribute between node i and j
+    
     Args:
         in_channels (int): Size of each input sample.
         out_channels (int): Size of each output sample.
-        bias (bool, optional): If set to :obj:`False`, the layer will not learn
+        bias (bool): If set to :obj:`False`, the layer will not learn
```
gcroci2, line 24:
```suggestion
        bias (bool, optional): If set to :obj:`False`, the layer will not learn
```

gcroci2 (changes requested):
Few comments:
- The `optional` keyword in the doc string (the one between round brackets) means that the argument has a default. I added some of them, but keep it in mind also for the other PRs. 
- Be sure that the spacing in the doc strings for the type is the same (e.g.`param(type1, optional): [...]` instead of `param (type1, optional): [...]`). Both ways are fine, just pick one and be consistent across all PRs and file. Also for this, be sure to have checked the spacing in all PRs. 
- For specifying the defaults in the doc strings, let's choose one way on doing it and let's modify everything accordingly. Options that work are: `Default: default_value.` or `Defaults to default_value`. Pick one and apply it to all files and PRs.

gcroci2 (changes requested):
Be sure to have checked usage of Optional type hinting. All the parameters that can have None (either as default or not default) should be hinted as Optional (see [typing.Optional](https://docs.python.org/3/library/typing.html#:~:text=On%20the%20other%20hand%2C%20if%20an%20explicit%20value%20of%20None%20is%20allowed%2C%20the%20use%20of%20Optional%20is%20appropriate%2C%20whether%20the%20argument%20is%20optional%20or%20not.%20For%20example%3A) docs). This is different from `optional` in the doc strings documentation though. The latter means that the argument is defaulted (to whathever value: int, None, float, str, ...).
So, for example, a parameter like `param: bool = True` will be in the docs as `param (bool, optional)` since it's defaulted to True, while a parameter like `param: Optional[bool]` will be in the docs as `param (bool)` since it's not defaulted so it's not optional (the `Optional` hint only means that it can be None). Finally, a parameter like `param: Optional[bool] = None` will be in the docs as `param (bool, optional)`, and it can be None, and it is defaulted to None. Please let me know if it's not clear :)

gcroci2 (approved):

gcroci2 (approved):

PR #362, opened 2023-02-15
perf: accelerate indexing of HDF5 files

joyceljy:
This PR accelerated the hdf5 indexing by:
1. skipped the function ```self._filter_targets``` when the variable ```self.target_filter``` is **None**.
2. Make the for loop of adding ```self.index_entries``` using python **list comprehension**.


deeprankcore/dataset.py
```diff
@@ -147,9 +147,10 @@ def _create_index_entries(self):
                     else:
                         entry_names = [entry_name for entry_name in self.subset if entry_name in list(hdf5_file.keys())]
 
-                    for entry_name in entry_names:
-                        if self._filter_targets(hdf5_file[entry_name]):
-                            self.index_entries += [(hdf5_path, entry_name)]
+                    #using list comprehension to skip _filter_targets method if target_filter is None
+                    self.index_entries += [(hdf5_path, entry_name) for entry_name in entry_names \
+                         if self.target_filter is None or self._filter_targets(hdf5_file[entry_name]) is True]
```
gcroci2, line 152:
```suggestion
                         if self._filter_targets(hdf5_file[entry_name])]
```

gcroci2 (changes requested):
Few notes:
- I see some edits (such as line 376 in dataset.py and lines 295-306 in utils/graph.py) which are on the main branch but appear on your PR. If you branched from main (after having pulled the last version of it doing `git pull`) this shouldn't be happening. 
- Please remember to delete deeprankcore/tools/plotgraph.py file, which is empty.
- List comprehension should speed up the process, but to be sure and to have some more precise timings you can use [cProfile](https://www.machinelearningplus.com/python/cprofile-how-to-profile-your-python-code/) module and then report the comparison between running `_create_index_entries` method before and after your changes. You could add such considerations on the [issue](https://github.com/DeepRank/deeprank-core/issues/340). To see an example about how to use the module, see [here](https://github.com/DeepRank/3D-Vac/blob/0372cadd9db7a695d901b6829ccc0fa31ccf8bcd/src/4_train_models/GNN/I/classification/struct/training.py#L237). You could enable cprofiler inside the `_create_index_entries` method and log the timings there before and after your changes, using the 100 samples example dataset I gave you.

joyceljy:
[cProfileTestResult.zip](https://github.com/DeepRank/deeprank-core/files/10793905/cProfileTestResult.zip)
I tested the performance time of the new index code using cProfile with 100 sample data.
From the cumulative time, there was a 2-sec improvement for per call in dataset.py.
The comparison result is in the attached file, the file name with _old is the performance result using the original code

gcroci2 (changes requested):
plotgraph.py needs to be in tools/, not tools/visualization/ :)

gcroci2:
> [cProfileTestResult.zip](https://github.com/DeepRank/deeprank-core/files/10793905/cProfileTestResult.zip) I tested the performance time of the new index code using cProfile with 100 sample data. From the cumulative time, there was a 2-sec improvement for per call in dataset.py. The comparison result is in the attached file, the file name with _old is the performance result using the original code
> 
> Cumulative time result after index accelerating: image
> 
> Cumulative time result before index accelerating: image

The improvement is of 0.002 s actually, right? From the screenshots I would say this. 
Last experiment you could do, is to pass multiple times the hdf5 file to the GraphDataset object in order to use more data points and see how this scales. I would advice to use ~1000 data points to have a more indicative measure of the improvement (with 100 we can't really see if it's significative)

joyceljy:
[IndexAccerlateTestResult.zip](https://github.com/DeepRank/deeprank-core/files/10804648/IndexAccerlateTestResult.zip)
Update: 
Tested 71000 samples of data using the new index accelerating method.
The performance improved by around 8 seconds in total.
Cumulative time result after index accelerating:
image

Cumulative time result before index accelerating:
image

deeprankcore/dataset.py
```diff
@@ -147,9 +147,13 @@ def _create_index_entries(self):
                     else:
                         entry_names = [entry_name for entry_name in self.subset if entry_name in list(hdf5_file.keys())]
 
-                    for entry_name in entry_names:
-                        if self._filter_targets(hdf5_file[entry_name]):
-                            self.index_entries += [(hdf5_path, entry_name)]
+                    #skip self._filter_targets when target_filter is None, improve performance using list comprehension.
+                    if (self.target_filter is None):
```
gcroci2, line 151:
```suggestion
                    if self.target_filter is None:
```

deeprankcore/dataset.py
```diff
@@ -147,9 +147,13 @@ def _create_index_entries(self):
...
+                        self.index_entries += [(hdf5_path, entry_name) for entry_name in entry_names]
```
gcroci2, line 152:
```suggestion
                        self.index_entries = [(hdf5_path, entry_name) for entry_name in entry_names]
```

deeprankcore/dataset.py
```diff
@@ -147,9 +147,13 @@ def _create_index_entries(self):
...
+                    else:
+                        self.index_entries += [(hdf5_path, entry_name) for entry_name in entry_names \
```
gcroci2, line 154:
```suggestion
                        self.index_entries = [(hdf5_path, entry_name) for entry_name in entry_names \
```

gcroci2 (changes requested):
Please commit my suggestions, then just to be sure ask also @DaniBodor a check. If all tests pass, then feel free to merge :)

DaniBodor:
> Please commit my suggestions, then just to be sure ask also @DaniBodor a check. I removed the "+" from the list comprehension since I think it's not needed. If all tests pass, then feel free to merge :)

I believe it should actually remain `+=` (as in the original), because we are adding to the list for each `entry_name`. Otherwise we keep overwriting the list and essentially only have indices for the final entry.

DaniBodor (approved):
Looks good, Joyce!
With your changes, line 130 (`self.index_entries = []`) has become obsolete. This was needed before, because we were adding entries to the list one by one in a loop, but now with the list comprehension it all happens at once.
You can delete this line and then (if all checks pass) you can go ahead and merge to main.

Also, could you give a 1 or 2 sentence summary (or bullet points) of your changes in the top comment of this PR. That can be very useful if in the future we want to look back at this and quickly know what was changed.

PR #369, opened 2023-02-23
test: split test contact.py

joyceljy:
I split the original test in features/test_contact.py into four smaller tests.
They were test_vanderwaals(), test_attractive_electrostatic(), test_repulsive_electrostatic() and test_cal_residue_contact().


DaniBodor (changes requested):
Thanks for picking this up, Joyce! It is much more clearly organized into groups the way you now split the tests up. 
Ideally, you could split them up even further into individual tests. Pretty much each time there's a comment block saying what is being tested, it could be a single test function. You can then use what is now in the comment as a docstring for that test.

Because most of the tests will be essentially similar, where mainly the inputs and assert statements will change, you can create a sort of helper function that contains the main body of the test, and then call this function with a different input each time.

Also, don't forget to tag which issue you will be closing with this PR ;)

Let me know (on Teams) if it's unclear what I mean with this or if you need help, and I can explain it better.

joyceljy:
Update:
I split the function into smaller test cases(based on the original comments). And also changed the comments into docstring format:) For the duplicate part appearing in each test cases, I defined a help function ``` _wrap_pdb_get_atomiccontact ```, which assigned the pdb_path and called the AtomicContact object. The parameters (residue_number & atom_name )needed for the AtomicContact object are passed from each test case.

DaniBodor (approved):
you can go ahead and merge :)

PR #370, opened 2023-02-24
docs: incorrect usage of Union in Optional type hints

DaniBodor:
In `DeeprankDataset` class, `Optional[Union[...]]` was used as a type hint in a few cases, where only one possibility existed (apart from the `None` of Optional).
This was probably a remnant of the previous formulation of `Union[..., None]`. However, because now `Optional` is used, `Union` should not be anymore.


joyceljy (approved):
Yes, thanks:)
This makes more sense because Union should contain two or more data type, and right here it seems to have one data type only. So Optional would be sufficient:)

PR #387, opened 2023-03-16
feat: add feature module for secondary structure

DTRademaker:
Secondary structure feature and test case. The feature extractor requires dssp to be installed (apt-get install dssp)


DaniBodor:
@cbaakman , can you help me figure out why DSSP is not running properly during the GitHub build workflow?

We have tried running it in 2 separate ways. Both work fine locally when running pytest, but then the build fails on GitHub

The original way Daniel wrote it gives a ValueError (see commit [a2d68a9](https://github.com/DeepRank/deeprank-core/pull/387/commits/a2d68a939610bf916b94bf5c81b01cd8f6893355), deeprankcore/features/secondary_structure.py:24: ValueError).
```
dssp_output = os.popen(f'dssp -i {pdb_path}').read()
dssp_lines = dssp_output[dssp_output.index('  #  RESIDUE'):].split('\n')[1:-1]
```
> ValueError: substring not found

I then tried doing it via Biopython, but that raised an exception from their own module (see latest [commit](https://github.com/DeepRank/deeprank-core/pull/387/commits/13e16a523b11a992548c36ce66a32243f5f6c231), deeprankcore/features/secondary_structure.py:54). I tried both `dssp='mkdssp'` and `dssp='dssp'` here, with the same result: an exception raised by the DSSP module of Biopython because it doesn't procude any output.
```
p = PDBParser(QUIET=True)
model = p.get_structure(Path(pdb_path).stem, pdb_path)[0]
dssp = DSSP(model, pdb_path, dssp='mkdssp')
```
> Exception: DSSP failed to produce an output

It does seem that DSSP is installing properly in the workflow action, but then doesn't actually work as intended when called upon in the code. Do you know why this could be or what is going wrong?

cbaakman:
What version of dssp did you install? Since 4.0 the software outputs a different format by default. That executable requires an additional parameter to make it output the original format.

DaniBodor:
> What version of dssp did you install? Since 4.0 the software outputs a different format by default. That executable requires an additional parameter to make it output the original format.
> 
> ```
> $ mkdssp --version
> mkdssp version 4.0.5
> $ mkdssp --help
> mkdssp [options] input-file [output-file]:
>   --dict arg             Dictionary file containing restraints for residues in 
>                          this specific target, can be specified multiple times.
>   --output-format arg    Output format, can be either 'dssp' for classic DSSP 
>                          or 'mmcif' for annotated mmCIF. The default is chosen 
>                          based on the extension of the output file, if any.
>   --min-pp-stretch arg   Minimal number of residues having PSI/PHI in range for
>                          a PP helix, default is 3
>   --write-other          If set, write the type OTHER for loops, default is to 
>                          leave this out
>   --components arg       Location of the components.cif file from CCD
>   --extra-compounds arg  File containing residue information for extra 
>                          compounds in this specific target, should be either in
>                          CCD format or a CCP4 restraints file
>   --mmcif-dictionary arg Path to the mmcif_pdbx.dic file to use instead of 
>                          default
>   -h [ --help ]          Display help message
>   --version              Print version
>   -v [ --verbose ]       verbose output
> ```

Thanks! I now notice that the workflow and my versions are different, so that probably explains it.

It seems that the workflow installed 4.0.4-1, although I didn't specify a version.
image

The workflow runs:
`sudo apt-get install -y dssp`

I now notice that locally I have 3.0.0 (even though I ran the same line a few days ago), so that could definitely explain the difference.

I'll look into making it work from here. 

PS: Do you actually have a recommended way to run dssp from a python script? I personally kind of like the Biopython implementation better than running it as an os command, but not sure if there's a better way. I just came across [this](https://anaconda.org/salilab/dssp), but haven't looked into it yet.

cbaakman:
I personally always run the subprocess.call and then parse the output file. But I think it would be cleaner to use the biopython implementation, using the dssp 4.0 executable.

DaniBodor:
Currently the secondary structure module works locally on any pdb files, but remotely (in GitHub actions) only on properly formatted PDB files (not the ones starting directly with ATOM lists).

I tried writing into the pdb files to fix this, but have not managed to pin down exactly which elements are needed to make it work. It is also unclear to me why there is a difference between local and remote. I suspect that on Snellius we will have same problem as on GitHub, but might be worth a test.

I have now adjusted the tests to run on pdb files that are correctly formatted, so that all tests pass. Also, I created a better exception, explaining the (likely) problem if it crashes.

I'm not sure whether this PR should remain open until we find a way to resolve it, or whether we should merge it in it's current form (it's usable, as long as pdb are formatted correctly). What do you think @gcroci2 ?

deeprankcore/features/secondary_structure.py
```diff
@@ -0,0 +1,150 @@
...
+        return t
+
+
+def _get_records(lines: List[str]):
+    seen = set()
+    seen_add = seen.add
+    return [x.split()[0] for x in lines if not (x in seen or seen_add(x))]
+
+
+def _check_pdb(pdb_path):
```
gcroci2, line 41: type hint missing here

tests/features/test_secondary_structure.py
```diff
@@ -0,0 +1,81 @@
...
+from deeprankcore.features.secondary_structure import (SecondarySctructure,
+                                                       _classify_secstructure,
+                                                       add_features)
+
+from . import build_testgraph
+
+
+def test_secondary_structure_residue():
+    test_case = '9api' # properly formatted pdb file
+    pdb_path = f"tests/data/pdb/{test_case}/{test_case}.pdb"
```
gcroci2, line 13: Here we could use PATH_TEST global tests variable to refer to "tests/" folder (`os.join(PATH_TEST, 'data', 'pdb', f'{test_case}', f'{test_case}.pdb')`)

tests/features/test_secondary_structure.py
```diff
@@ -0,0 +1,81 @@
...
+        ), f'Ground truth examples: res {res[1]} {res[0]} is not {(res[2])}.'
+        assert np.all(
+            [np.array_equal(node_info[2], res[3].onehot)
+                for node_info in node_list]
+        ), f'Ground truth examples: res {res[1]} {res[0]} is not {res[3]}.'
+
+
+def test_secondary_structure_atom():
+    test_case = '1ak4' # ATOM list
+    pdb_path = f"tests/data/pdb/{test_case}/{test_case}.pdb"
```
gcroci2, line 55: Same as above for using PATH_TEST

tests/features/test_secondary_structure.py
```diff
@@ -0,0 +1,81 @@
...
+    
+    # Create a list of node information (residue number, chain ID, and secondary structure features)
+    node_info_list = [[node.id.residue.number, 
+                       node.id.residue.chain.id, 
+                       node.features[Nfeat.SECSTRUCT]] 
+                            for node in graph.nodes]
+
+    # check entire DSSP file
+    # residue number @ pos 5-10, chain_id @ pos 11, secondary structure @ pos 16
+    with open(f'tests/data/dssp/{test_case}.dssp.txt', encoding="utf8") as file:
```
gcroci2, line 67: Same as above for using PATH_TEST, also using os.join (system independent) is preferable than using `/`

tests/test_integration.py
```diff
@@ -18,7 +18,14 @@
 from deeprankcore.utils.exporters import HDF5OutputExporter
 from deeprankcore.utils.grid import GridSettings, MapMethod
 
-from . import PATH_TEST
+pdb_path = str("tests/data/pdb/3C8P/3C8P.pdb")
```
gcroci2, line 21: Same as other comments here, I think is better to use PATH_TEST variable and os.join() method for defining paths

gcroci2 (approved):
Great addition, thank you all :) I only added minor annoying comments about how we define paths in the tests, since I noticed with this PR that we're not doing it in the advised way. I didn't comment them all.
We can either:
- Use forward slashes (`/`) or commas (`,`) with `pathlib` functions - both options work. The `Path()` object will convert forward slashes or commas into the correct kind of slash for the current operating system (nice blog post about it [here](https://medium.com/@ageitgey/python-3-quick-tip-the-easy-way-to-deal-with-file-paths-on-windows-mac-and-linux-11a072b58d5f#:~:text=You%20should%20use%20forward%20slashes,operator%20directly%20in%20your%20code.)).
- Use `os.path.join()` method.

Apparently, `pathlib` is considered more modern, but it seems to be more a matter of taste than functionality (see for example this [stackoverflow thread](https://stackoverflow.com/questions/67112343/pathlib-vs-os-path-join-in-python)). We can agree on one of the two ways and when we'll add edits to the tests we'll stick with it.

DaniBodor:
> Great addition, thank you all :) I only added minor annoying comments about how we define paths in the tests, since I noticed with this PR that we're not doing it in the advised way. I didn't comment them all. We can either:
> 
>     * Use forward slashes (`/`) or commas (`,`) with `pathlib` functions - both options work. The `Path()` object will convert forward slashes or commas into the correct kind of slash for the current operating system (nice blog post about it [here](https://medium.com/@ageitgey/python-3-quick-tip-the-easy-way-to-deal-with-file-paths-on-windows-mac-and-linux-11a072b58d5f#:~:text=You%20should%20use%20forward%20slashes,operator%20directly%20in%20your%20code.)).
> 
>     * Use `os.path.join()` method.
> 
> 
> Apparently, `pathlib` is considered more modern, but it seems to be more a matter of taste than functionality (see for example this [stackoverflow thread](https://stackoverflow.com/questions/67112343/pathlib-vs-os-path-join-in-python)). We can agree on one of the two ways and when we'll add edits to the tests we'll stick with it.


Shall we open a new issue for this to do it consistently throughout all test files rather than fixing a few examples here? 
Not knowing the best practices, I was actually getting rid of the global variable in the few cases where it was used in another PR I was working on, as to me it seemed less clear that way than using hard coded paths.
I also think we can do with decreasing the folder structure by 1, as I don't see the point of having files in (nearly) identically named folders.

deeprankcore/features/secondary_structure.py
DaniBodor, line 41:
```suggestion
def _check_pdb(pdb_path: str):
```

PR #396, opened 2023-03-19
docs: create tables for feature documentation

DaniBodor:
Moved feature descriptions into a table format. 

Not sure this will parse as expected in the readthedocs page. How can I test that?


DaniBodor:
replaced by #419

PR #400, opened 2023-03-20
refactor: make feature modules and tests uniform and ditch duplicate code

DaniBodor:
After looking at the different features and their tests a lot recently, i got really annoyed about them all being formatted different while doing the same thing. So decided to make it nice instead


gcroci2 (approved):
Much more in order now <3

PR #401, opened 2023-03-20
fix: match the assumed aa order of the code to the pssm files

DaniBodor:
There should be some form of documentation on what the pssm files should look like, as it is very easy get it wrong and very difficult to diagnose or even detect if it is.

Even nicer would be to read the order from the top of the pssm file instead of assuming a fixed order.

Also, ideally if pssm files are not present, the code shouldn't crash but just skip that feature and throw a warning (is nicer for users that don't have pssm files not to have to dive into code to actively remove the feature).

### DIFF
--- a/deeprankcore/features/conservation.py
+++ b/deeprankcore/features/conservation.py
@@ -7,13 +7,13 @@ from deeprankcore.molstruct.atom import Atom
 from deeprankcore.utils.graph import Graph
 from deeprankcore.domain import nodestorage as Nfeat
 
-profile_amino_acid_order = sorted(amino_acids, key=lambda aa: aa.one_letter_code)
-
 
 def add_features( # pylint: disable=unused-argument
     pdb_path: str, graph: Graph,
     single_amino_acid_variant: Optional[SingleResidueVariant] = None
     ):
+
+    profile_amino_acid_order = sorted(amino_acids, key=lambda aa: aa.three_letter_code)
 
     for node in graph.nodes:
         if isinstance(node.id, Residue):
--- a/deeprankcore/utils/parsing/pssm.py
+++ b/deeprankcore/utils/parsing/pssm.py
@@ -4,10 +4,6 @@ from deeprankcore.molstruct.residue import Residue
 from deeprankcore.utils.pssmdata import PssmRow, PssmTable
 from deeprankcore.domain.aminoacidlist import amino_acids
 
-amino_acids_by_letter = {
-    amino_acid.one_letter_code: amino_acid for amino_acid in amino_acids
-}
-
 
 def parse_pssm(file_: TextIO, chain: Chain) -> PssmTable:
     """Read the PSSM data.
@@ -20,6 +16,7 @@ def parse_pssm(file_: TextIO, chain: Chain) -> PssmTable:
         PssmTable: The position-specific scoring table, parsed from the pssm file.
     """
 
+    amino_acids_by_letter = {amino_acid.one_letter_code: amino_acid for amino_acid in amino_acids}
     conservation_rows = {}
 
     # Read the pssm header.
--- a/deeprankcore/utils/pssmdata.py
+++ b/deeprankcore/utils/pssmdata.py
@@ -1,10 +1,6 @@
 from typing import Optional, Dict, List
 from deeprankcore.molstruct.aminoacid import AminoAcid
-from deeprankcore.domain.aminoacidlist import amino_acids
 
-amino_acids_by_letter = {
-    amino_acid.one_letter_code: amino_acid for amino_acid in amino_acids
-}
 
 class PssmRow:
     """Holds data for one position-specific scoring matrix row."""
--- a/tests/data/pssm/1ATN/1ATN.A.pdb.deeprank.pssm
+++ b/tests/data/pssm/1ATN/1ATN.A.pdb.deeprank.pssm
@@ -1,1 +1,0 @@
-pdbresi pdbresn seqresi seqresn    A    R    N    D    C    Q    E        G    H    I    L    K    M    F    P    S    T    W    Y    V   IC


DaniBodor:
this PR could def use a shorter name.

gcroci2 (commented):
Nice checks :-)

I'd avoid adding `exclude_feature_module` (the least parameters the better), and I'd implement the solution we were discussing, which is defaulting the modules to the essentials and making `feature_modules` required. I don't think opening a new PR for this is needed, right?

.github/actions/install-python-and-package/action.yml
```diff
@@ -60,7 +60,7 @@ runs:
         # Only way to install msms is through conda
         conda install -c bioconda msms
         # Safest way to install PyTorch and PyTorch Geometric is through conda
-        conda install pytorch torchvision torchaudio cpuonly -c pytorch
+        conda install pytorch torchvision torchaudio cpuonly -c pytorch==2.0.0
```
gcroci2: Great! I've also just noticed that in one of my branches. Also I had a question: why is the line 66 needed? (# Install optional pyg dependencies) Shouldn't `conda install pyg -c pyg` take care of the pyg dependencies?

gcroci2 (approved):

README.md
```diff
@@ -59,7 +59,9 @@ Before installing deeprankcore you need to install:
  * [msms](https://ssbio.readthedocs.io/en/latest/instructions/msms.html): `conda install -c bioconda msms`. *For MacOS with M1 chip users*: you can follow [these instructions](https://ssbio.readthedocs.io/en/latest/instructions/msms.html).
  * [dssp](https://swift.cmbi.umcn.nl/gv/dssp/): `sudo apt-get install dssp`
     * See [DSSP docs](https://ssbio.readthedocs.io/en/latest/instructions/dssp.html) for installing it on Mac OSX
- * [pytorch](https://pytorch.org/get-started/locally/): `conda install pytorch torchvision torchaudio cpuonly -c pytorch` or `conda install pytorch torchvision torchaudio pytorch-cuda=11.7 -c pytorch -c nvidia`, for taking advantage of GPUs.
+ * [pytorch](https://pytorch.org/get-started/locally/):
```
gcroci2, line 62: Please change the same in the getstarted file in docs

deeprankcore/query.py
```diff
@@ -236,11 +280,12 @@ def process( # pylint: disable=too-many-arguments, too-many-locals
...
-                Each feature's module must implement the :py:func:`add_features` function, and features' modules can be found (or should be placed
-                in case of a custom made feature) in `deeprankcore.features` folder. 
-                Defaults to None, which means that all available modules in `deeprankcore.features` are used to generate
-                the features. 
+            feature_modules (Optional[Union[ModuleType, List[ModuleType], str, List[str]]], optional): Features' module or list of features' modules
+                used to generate features (given as string or as an imported module). Each module must implement the :py:func:`add_features` function, 
+                and features' modules can be found (or should be placed in case of a custom made feature) in `deeprankcore.features` folder. 
+                Use 'all' to include all feature modules.
+                If set to None or 'all', all available modules in `deeprankcore.features` are used to generate the features.
+                Defaults to None, which will only use the basic feature modules deeprankcore.features.components and deeprankcore.features.contact.
```
gcroci2, line 288: Instead of defaulting to None, I would default it to deeprankcore.features.components and deeprankcore.features.contact explicitely

deeprankcore/query.py
```diff
@@ -236,11 +280,12 @@ def process( # pylint: disable=too-many-arguments, too-many-locals
...
-            feature_modules(Union[ModuleType, List[ModuleType]], optional): Features' module or list of features' modules used to generate features. 
-                Each feature's module must implement the :py:func:`add_features` function, and features' modules can be found (or should be placed
-                in case of a custom made feature) in `deeprankcore.features` folder. 
-                Defaults to None, which means that all available modules in `deeprankcore.features` are used to generate
-                the features. 
+            feature_modules (Optional[Union[ModuleType, List[ModuleType], str, List[str]]], optional): Features' module or list of features' modules
+                used to generate features (given as string or as an imported module). Each module must implement the :py:func:`add_features` function, 
+                and features' modules can be found (or should be placed in case of a custom made feature) in `deeprankcore.features` folder. 
+                Use 'all' to include all feature modules.
+                If set to None or 'all', all available modules in `deeprankcore.features` are used to generate the features.
```
gcroci2, line 287: I would say that only if set to 'all' all the module will be used

deeprankcore/query.py
DaniBodor: I tried this, but the linter didn't like it. Should I suppress the linter?

gcroci2 (commented):
I added comments about minor edits in the `process` method. I think that explicitly defaulting the modules (not using None) and allowing for 'all' but not None is clear from the user perspective.

.github/actions/install-python-and-package/action.yml
DaniBodor: Don't know why, but it doesn't

deeprankcore/query.py
gcroci2: Which error did it give?

deeprankcore/query.py
DaniBodor: Yeah, this was a typo.

deeprankcore/query.py
DaniBodor: [dangerous-default-value](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/dangerous-default-value.html), basically it doesn't like lists as defaults

DaniBodor:
> I added comments about minor edits in the `process` method. I think that explicitly defaulting the modules (not using None) and allowing for 'all' but not None is clear from the user perspective.

Agreed!

gcroci2 (approved):

PR #404, opened 2023-03-23
style: add .vscode settings.json

DaniBodor:
When using VSCode:
- automatically sort imports (using isort) upon saving a .py file
- flag prospector problems upon typing

NOTE: build is failing due to torch_scatter not being automatically installed with pyg anymore. This is resolved in #406


gcroci2 (approved):
I checked only the settings.json and few files, but if the tests do not fail we can conclude that it works properly. Very handy addition :)

PR #418, opened 2023-04-18
feat: Improvement on Feature Standardization

joyceljy:
Improvement on feature standardization by applying suitable transformation for different features.
For now, we will modify the package by adding a dictionary indicating the type of transformation in which the feature should apply. The detailed description are as follows:

- We won't normalize nor standardize one-hot encoded features. For motivation, see for example [this](https://datascience.stackexchange.com/questions/56444/do-i-need-to-standardize-my-one-hot-encoded-labels#:~:text=Among%20the%20features%2C%20it%20has,%2Dstyle%20or%20engine%2Dtype.) thread (this is the common opinion among the community).
- In general, the rule of thumb for deciding when applying transformation before standardization is to have a distribution that widespreads the features values, ideally resembling a Gaussian, but not mandatorily (for example `electrostatic` and `vanderwaals` cube version is better than the original one). 
- Features to which we won't apply log, but standardization directly: `res_size`, `res_charge`, `hb_donors`, `hb_acceptors`, `hse`, `irc_` features, `res_mass`, `res_pI`, `distance`
- Features to which we'll apply log(x+1): `res_depth`, `bsa`, `info_content`
- Features to which we'll apply square root: `sasa`
- Features to which we'll apply cube root: `electrostatic`, `vanderwaals`
- We'll remove for now `pssm` since it's not correctly computed

Further steps:
Implement a user-defined dictionary, the user can decide which transformation they want to apply to each feature and insert into the dictionary.


DaniBodor:
> We'll remove for now `pssm` since it's not correctly computed

it is a one-hot encoded feature, so wouldn't need to be touched anyway.

deeprankcore/dataset.py
```diff
@@ -618,25 +619,28 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
...
                 Defaults to True.
 
-            dataset_train: (class:`GraphDataset`, optional): if train is True, assign here the training set. This parameter is considered only if standardize
-                flag is set to True.
+            dataset_train (class:`GraphDataset`, optional): if train is True, assign here the training set.
+                Defaults to None.
+                
+            features_transform (Optional[dict], optional): Dictionary to determine the transformation methods and whether to apply
+                standardization to each feature in the dictionary.
+                For example: features_transform = {'bsa':{'transform':lambda t:np.log(t+1),'standardize':True}} for feature bsa.
```
gcroci2, line 630:
```suggestion
                For example: features_transform = {'bsa': {'transform': lambda t:np.log(t+1),' standardize': True}} for the feature `bsa`.
```

deeprankcore/dataset.py
```diff
@@ -618,25 +619,28 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
...
+                A "all" key can be set in the dictionary, which performs the same standardize and transform for all features,
+                for example: features_transform = {'all':{'transform':lambda t:np.log(t+1),'standardize':True}}.
```
gcroci2, line 632:
```suggestion
                for example: features_transform = {'all': {'transform': lambda t:np.log(t+1), 'standardize': True}}.
```

tests/test_dataset.py
```diff
@@ -392,20 +455,20 @@ def test_graph_standardization_logic(self):
...
         )
-
+        # mean and devs should be None
         assert dataset_train.means == dataset_test.means
         assert dataset_train.devs == dataset_test.devs
+        assert dataset_train.means is None
+        assert dataset_train.devs is None
 
         # raise error if dataset_train is not provided
         with self.assertRaises(TypeError):
```
gcroci2, line 468: This and the `with self.assertRaises(TypeError):` below (row 476) should be moved to another test, we can call it `test_logic_graphdataset` and we can place it above `test_standardization_logic_graphdataset`

tests/test_dataset.py
```diff
@@ -14,23 +15,47 @@
 
 node_feats = [Nfeat.RESTYPE, Nfeat.POLARITY, Nfeat.BSA, Nfeat.RESDEPTH, Nfeat.HSE, Nfeat.INFOCONTENT, Nfeat.PSSM]
 
+def _cal_mean_std( # noqa: MC0001, pylint: disable=too-many-locals
+                       hdf5_path: str,
+                       features_transform:dict,
+                       feat:str
```
gcroci2, line 21:
```suggestion
                       feat: str
```

tests/test_dataset.py
```diff
@@ -14,23 +15,47 @@
...
+                
+        df_dict[feat] = [
+            f[entry_name][Nfeat.NODE][feat][:]
+            if f[entry_name][Nfeat.NODE][feat][()].ndim == 1
+            else f[entry_name][Nfeat.NODE][feat][()] for entry_name in entry_names]
+        #apply transformation
+        if transform:
+            df_dict[feat]=[transform(row) for row in df_dict[feat]]
+        
+        df = pd.DataFrame(data=df_dict)
```
gcroci2, line 41: We don't need a pandas df here, since we're calculating the mean and the std dev for a single feature `feat`. You can safely remove `df_dict` and `df` variables. Instead, you can just calculate what now is `df_dict[feat]` as something like `feat_values` in the form of a list (same way as it is now) which then can be transformed if needed and finally converted into a numpy array for computing the mean and the std dev.

tests/test_dataset.py
```diff
@@ -14,23 +15,47 @@
...
+ 
+        means = {col: round(np.concatenate(df[col].values).mean(), 1) if isinstance(df[col].values[0], np.ndarray) \
```
gcroci2, line 43: means is actually a single mean here, so I suggest to change the name to `mean`. Same for devs below, change it to `dev`. You'll have something like:
```python
round(np.concatenate(feat_values.mean(), 1) if isinstance(feat_values[0], np.ndarray) else round(np.array(feat_values).mean(), 1)
```

tests/test_dataset.py
```diff
@@ -14,23 +15,47 @@
...
+            else round(df[col].values.mean(), 1) \
+            for col in df.columns[1:]}
+        devs = {col: round(np.concatenate(df[col].values).std(), 1) if isinstance(df[col].values[0], np.ndarray) \
+            else round(df[col].values.std(), 1) \
+            for col in df.columns[1:]}
+        
+        means_devs=[means,devs]
+
+        return means_devs
```
gcroci2, line 52: Returning two variables is much cleaner, something like
```python
return mean, dev
```

tests/test_dataset.py
```diff
@@ -420,11 +483,31 @@ def test_graph_standardization_logic(self):
...
+        dataset_test_transform = GraphDataset(
+            hdf5_path = hdf5_path,
+            target='binary',
+            node_features = [node_feat_test],
+            edge_features =[],
+            features_transform=features_transform
+        )
+        
+        #calculate transformation and standardization maunally
+        means_devs=_cal_mean_std(hdf5_path,features_transform,node_feat_test)
```
gcroci2, line 505: Following on my comments above about `_cal_mean_std`, here you'll have:
```python
mean, dev = _cal_mean_std(hdf5_path, features_transform, node_feat_test)
```

Remember to modify the asserts below as well.

Also note the spaces after punctuation like `,` to make the writing less compact and more clearer.

tests/test_dataset.py
```diff
@@ -310,80 +361,92 @@ def test_graph_standardize(self):
...
         )
-
+        
+        assert dataset_train.features_transform == dataset_test.features_transform
         assert dataset_train.means == dataset_test.means
         assert dataset_train.devs == dataset_test.devs
 
-        # without specifying standardization in training set
+        # features_transform setted in train
+        dataset_train = GraphDataset(
```
gcroci2, line 431: No need to redefine it here, you can safely delete it

tests/test_dataset.py
```diff
@@ -310,80 +361,92 @@ def test_graph_standardize(self):
...
+                        assert 0.8 < dev < 1.2
 
-    def test_graph_standardization_logic(self):
+    def test_standardization_logic_graphdataset(self):
 
         hdf5_path = "tests/data/hdf5/train.hdf5"
-
-        # normal logic 
+        features_transform={'all':{'transform':None,'standardize':True}}
+        features_transform_nostandardize={'all':{'transform':None,'standardize':False}}
```
gcroci2, line 410: Find a shorter name, like `other_feature_transform`

tests/test_dataset.py
```diff
@@ -420,11 +483,31 @@ def test_graph_standardization_logic(self):
...
+    def test_feature_transform_mean_std(self):
+        node_feat_test='bsa'
+        hdf5_path = "tests/data/hdf5/train.hdf5"
+        features_transform={'bsa':{'transform':lambda t:np.log(t+1),'standardize':True}}
+        
+        #calculate transformation and standardization form GraphDataset object
+        dataset_test_transform = GraphDataset(
+            hdf5_path = hdf5_path,
+            target='binary',
+            node_features = [node_feat_test],
```
gcroci2, line 499:
```suggestion
            node_features = node_feat_test,
```

tests/test_dataset.py
```diff
@@ -288,15 +334,20 @@ def test_graph_save_hist(self):
         assert len(os.listdir(output_directory)) > 0
 
         rmtree(output_directory)
-
-    def test_graph_standardize(self):
+    
+    def test_standardize_graphdataset(self):# noqa: MC0001, pylint: disable=too-many-locals
```
gcroci2, line 338: Please add also a test called like `test_standardize_all_graphdataset` which tests standardization on all features, using a dict with `'all': {standardize: True}`

tests/test_dataset.py
```diff
@@ -310,80 +361,92 @@ def test_graph_standardize(self):
...
-
-                mean = values.mean()
-                dev = values.std()
-
-                assert -0.2 < mean < 0.2
-                assert 0.8 < dev < 1.2
+            for key, values in features_dict.items():
+                if(key in features_transform):
+                    standardization = features_transform.get(key, {}).get('standardize')
+                    if standardization: #Feature contains in dictionary & standardization=True
```
gcroci2, line 400: here you're not testing `hb_donors`, for which standardize is False. In order to test it, you can define another GraphDataset like `no_std_dataset` without defining a `features_transform` dict, and then you can verify:
1) that features of `dataset` which are in `features_transform` with standardize True have mean and dev as indicated (mean around 0 and dev around 1) and different mean and dev from the same features in `no_std_dataset`
2) that features of `dataset` which are in `features_transform` with standardize False or which are not in the dict have mean and dev equal to the ones in `no_std_dataset`

tests/test_dataset.py
```diff
@@ -288,15 +334,20 @@ def test_graph_save_hist(self):
...
+        
         dataset = GraphDataset(
             hdf5_path = "tests/data/hdf5/train.hdf5",
-            target='binary',
-            standardize=True
+            target = 'binary',
+            features_transform = features_transform
         )
 
         with h5py.File(hdf5_path, 'r') as f5:
```
gcroci2, line 353: This with statement until the end of `features_dict` population can be wrapped up in an internal function called like `get_features(hdf5_path, dataset)`; it will be more readable this way and we'll use this function for testing the transformations as well.

gcroci2 (changes requested):
Very useful and nice changes :) I left minor comments, please ask my review again once you're done, we're almost there! Also in general, please leave a space before `=` symbol and after it, and also a space after punctuation like `,` and `:`

We'll look at the PR together once you implement these changes and we'll finalize the following:
- In `test_standardize_graphdataset`, we need to test `hb_donors`, for which standardize is False. And in general all the rest of the features not indicated in the dict. In order to test it, we can use `_cal_mean_std` with `hb_donors` feature; in general, we need to verify that features of `dataset` which are in `features_transform` with standardize True have mean and dev as indicated (mean around 0 and dev around 1) and different mean and dev from the same features not touched (maybe using `_cal_mean_std`); we also need to verify that features of `dataset` which are in `features_transform` with standardize False or which are not in the dict have mean and dev equal to the ones not touched, again maybe using `_cal_mean_std`
- `test_feature_transform_mean_std` partially tests the transformation, so we need to implement a smart way to really test transformations

tests/test_dataset.py
```diff
@@ -420,11 +483,31 @@ def test_graph_standardization_logic(self):
...
+        node_feat_test='bsa'
+        hdf5_path = "tests/data/hdf5/train.hdf5"
+        features_transform={'bsa':{'transform':lambda t:np.log(t+1),'standardize':True}}
+        
+        #calculate transformation and standardization form GraphDataset object
+        dataset_test_transform = GraphDataset(
+            hdf5_path = hdf5_path,
+            target='binary',
+            node_features = [node_feat_test],
+            edge_features =[],
```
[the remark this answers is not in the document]
joyceljy: From the document of GraphDataSet, we cannot pass the edge_features as None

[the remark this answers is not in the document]
gcroci2: From the docs, it is an [Optional parameter](https://docs.python.org/3/library/typing.html#:~:text=type%20expressions.-,typing.Optional%C2%B6,-Optional%20type.), which means that it can be None :)

joyceljy:
> Very useful and nice changes :) I left minor comments, please ask my review again once you're done, we're almost there! Also in general, please leave a space before `=` symbol and after it, and also a space after punctuation like `,` and `:`
> 
> We'll look at the PR together once you implement these changes and we'll finalize the following:
> 
> * In `test_standardize_graphdataset`, we need to test `hb_donors`, for which standardize is False. And in general all the rest of the features not indicated in the dict. In order to test it, we can use `_cal_mean_std` with `hb_donors` feature; in general, we need to verify that features of `dataset` which are in `features_transform` with standardize True have mean and dev as indicated (mean around 0 and dev around 1) and different mean and dev from the same features not touched (maybe using `_cal_mean_std`); we also need to verify that features of `dataset` which are in `features_transform` with standardize False or which are not in the dict have mean and dev equal to the ones not touched, again maybe using `_cal_mean_std`
> * `test_feature_transform_mean_std` partially tests the transformation, so we need to implement a smart way to really test transformations

I think not all the features with standardized True would got mean around 0 and dev around 1 even after transformation. For example for feature sasa, before transformation, its mean=45 & dev=41.5, and after transformation its mean=5.7 & dev=3.5.

gcroci2:
> > Very useful and nice changes :) I left minor comments, please ask my review again once you're done, we're almost there! Also in general, please leave a space before `=` symbol and after it, and also a space after punctuation like `,` and `:`
> > We'll look at the PR together once you implement these changes and we'll finalize the following:
> > 
> > * In `test_standardize_graphdataset`, we need to test `hb_donors`, for which standardize is False. And in general all the rest of the features not indicated in the dict. In order to test it, we can use `_cal_mean_std` with `hb_donors` feature; in general, we need to verify that features of `dataset` which are in `features_transform` with standardize True have mean and dev as indicated (mean around 0 and dev around 1) and different mean and dev from the same features not touched (maybe using `_cal_mean_std`); we also need to verify that features of `dataset` which are in `features_transform` with standardize False or which are not in the dict have mean and dev equal to the ones not touched, again maybe using `_cal_mean_std`
> > * `test_feature_transform_mean_std` partially tests the transformation, so we need to implement a smart way to really test transformations
> 
> I think not all the features with standardized True would got mean around 0 and dev around 1 even after transformation. For example for feature sasa, before transformation, its mean=45 & dev=41.5, and after transformation its mean=5.7 & dev=3.5. Do you have any suggestions on setting up the mean and dev range? For now I do something like this:
> 
> ```
> for key, values in features_dict.items():
>             if key in features_transform: 
>                 transform = features_transform.get(key, {}).get('transform')
>                 means = []
>                 devs = []
>                 (mean, dev) = _cal_mean_std(hdf5_path, features_transform, key)
>                 means.append(mean)
>                 devs.append(dev)
>                 means.append(values.mean())
>                 devs.append(values.std())
>                 if transform: #test transformed features
>                     assert means[0] != means[1]
>                     assert devs[0] != devs[1] 
>                     assert -10 < means[0] < 10
>                     assert -5 < devs[0] < 5
>                 else: #test hb_doners, no transformation so mean & std remain the same.
>                     assert means[0] == means[1]
>                     assert devs[0] == devs[1]
> ```

Another way to verify that you're actually standardizing, is to do the inverse calculation (destandardizing the values) and verifying that the mean and the std dev of these back-transformed values are the same as before. 

So in terms of code it would be, `values` being the ones after transformation (if present) and standardization: `values_no_std = values * dev + mean`
Then you can test that `values_no_std.mean()`  and `values_no_std.std()` are the same as the ones obtained with the feature not standardized.

gcroci2 (approved):

PR #420, opened 2023-05-02
test: improve `utils/test_graph.py` module

gcroci2:
Originally we were testing everything in only one test; now I divided the tests into three more logical ones, and I added few more asserts to improve coverage.


DaniBodor (approved):
Much more structured now :sparkles:.

I feel that there's a fair amount of duplicate code between the last 2 tests, so wonder if it's really necessary to separate those. Not a big deal, though, so if you don't want to spend more time on it, it's fine to leave as is as well.

PR #427, opened 2023-05-22
fix: epoch number in `test()` and test on the correct model

gcroci2:
Main changes proposed with this PR:
- `train` method has two distinct parameters now for indicating if the user wants to save either the best or the last model, and if the model needs to be saved. This way it's easier to do internal checks, and it's clearer for the user. 
- `save_model` method it's internal now and it saves only the state dict by default, into a variable which is then used in the code
  - The state dict needs to be retained to load the states of the best/last model at the end of the training loop
  - After the training loop is over, the best/last model state is saved in a file if the new flag `save_model` is True
- Before these edits, we were overriding the file every time we were saving the best model. Also, we were not loading the state dicts of the best model to the self.model and self.optimizer attributes. So we were then testing the model using the last updated model, which was always the last one. But in cases in which `best_model` was selected, again last was run on test set. Now it should be fixed.
- I modified the call to early stopping into the training loop in order to compare the current validation loss with the current training loss (and not the minimum training loss); indeed in each epoch we're interested in evaluating the current gap between the training and the validation loss. 
- I moved the early stopping call into the `if validate` statement, because if there is no validation there is no early stopping call. 
- We were saving test data into the exporter using epoch 0 as a placeholder; now the epoch saved is the one in which the model's state was saved, and that was actually run on the test, so the best/last one according to the user's choice. It's less confusing this way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,12 +290,11 @@ trainer.configure_optimizers(torch.optim.Adamax, lr = 0.001, weight_decay = 1e-0
 
 ```
 
-Then the Trainer can be trained and tested, and the model can be saved:
+Then the Trainer can be trained and tested; the best model in terms of validation loss is saved by default, and the user can modify so or indicate where to save it using the `train()` method parameters `save_model` and `output_prefix`.
 
 ```python
 trainer.train(nepoch = 50, batch_size = 64, validate = True)
 trainer.test()
-trainer.save_model(filename = "<output_model_path.pth.tar>")
 
 ```
 
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ trainer.configure_optimizers(torch.optim.Adamax, lr = 0.001, weight_decay = 1e-0
 
 ```
 
-Then the Trainer can be trained and tested; the best model in terms of validation loss is saved by default, and the user can modify so or indicate where to save it using the `train()` method parameters `save_model` and `output_prefix`.
+Then the Trainer can be trained and tested; the best model in terms of validation loss is saved by default, and the user can modify so or indicate where to save it using the `train()` method parameters `save_model` and `filename`.
 
 ```python
 trainer.train(nepoch = 50, batch_size = 64, validate = True)
--- a/README.md
+++ b/README.md
@@ -290,10 +290,14 @@ trainer.configure_optimizers(torch.optim.Adamax, lr = 0.001, weight_decay = 1e-0
 
 ```
 
-Then the Trainer can be trained and tested; the best model in terms of validation loss is saved by default, and the user can modify so or indicate where to save it using the `train()` method parameters `save_model` and `filename`.
+Then the Trainer can be trained and tested; the best model in terms of validation loss is saved by default, and the user can modify so or indicate where to save it using the `train()` method parameter `filename`.
 
 ```python
-trainer.train(nepoch = 50, batch_size = 64, validate = True)
+trainer.train(
+    nepoch = 50,
+    batch_size = 64,
+    validate = True,
+    filename = "<my_folder/model.pth.tar>")
 trainer.test()
 
 ```
--- a/deeprankcore/trainer.py
+++ b/deeprankcore/trainer.py
@@ -641,7 +641,7 @@ class Trainer():
                     valid_losses.append(loss_)
                     if best_model:
                         if min(valid_losses) == loss_:
-                            checkpoint_model = self.save_model()
+                            checkpoint_model = self._save_model()
                             self.epoch_saved_model = epoch
                             _log.info(f'Best model saved at epoch # {self.epoch_saved_model}.')
                     # check early stopping criteria (in validation case only)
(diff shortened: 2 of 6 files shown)


joyceljy (approved):

deeprankcore/trainer.py
```diff
@@ -518,8 +518,9 @@ def train( # pylint: disable=too-many-arguments, too-many-branches, too-many-loc
         min_epoch: int = 10, 
         validate: bool = False,
         num_workers: int = 0,
-        save_best_model: Optional[bool] = True,
-        output_prefix: Optional[str] = None,
+        best_model: bool = True,
+        save_model: bool = True,
+        filename: str = 'model.pth.tar'
```
DaniBodor, line 523: How about reducing the number of parameters by allowing the filename to be optional and if None --> don't save?


```suggestion
        filename: Optional[str] = 'model.pth.tar'
        best_model: bool = True,
```

deeprankcore/trainer.py
```diff
@@ -541,13 +542,15 @@ def train( # pylint: disable=too-many-arguments, too-many-branches, too-many-loc
...
+                        If True, the best model (in terms of validation loss) is selected for later testing or saving.
+                        If False, the last model tried is selected.
                         Defaults to True.
-            output_prefix (Optional[str], optional): Name under which the model is saved. A description of the model settings is appended to the prefix.
-                        Defaults to None.
+            save_model (bool, optional): 
+                        If True, the selected model is saved in <filename>.
+                        If False, no model is saved.
+                        Defaults to True.
+            filename (str, optional): Name of the file where to save the selected model. Defaults to 'model.pth.tar'.
```
DaniBodor, line 553: See comment above

```suggestion
            filename (Optional[str, optional]): Name of the file where to save the selected model. 
                ```Defaults to 'model.pth.tar'.
                If None, model is not saved
            best_model (bool, optional): 
                        If True, the best model (in terms of validation loss) is selected for later testing or saving.
                        If False, the last model tried is selected.
                        Defaults to True.
            save_model (bool, optional): 
                        If True, the selected model is saved in <filename>.
                        If False, no model is saved.
                        Defaults to True.
```

deeprankcore/trainer.py
```diff
@@ -892,12 +900,12 @@ def _load_params(self):
...
 
-    def save_model(self, filename='model.pth.tar'):
+    def _save_model(self, filename: Optional[str] = None):
         """
         Saves the model to a file.
 
         Args:
-            filename (str, optional): Name of the file. Defaults to 'model.pth.tar'.
+            filename (str, optional): Name of the file. Defaults to None.
         """
```
DaniBodor, line 909: this no longer requires an argument

docs/getstarted.md
```diff
@@ -268,12 +268,11 @@ trainer.configure_optimizers(torch.optim.Adamax, lr = 0.001, weight_decay = 1e-0
 
 ```
 
-Then the Trainer can be trained and tested, and the model can be saved:
+Then the Trainer can be trained and tested; the best model in terms of validation loss is saved by default, and the user can modify so or indicate where to save it using the `train()` method parameters `save_model` and `filename`.
```
DaniBodor, line 271: I would add arguments for this to the code snippet below

README.md
```diff
@@ -290,12 +290,11 @@ trainer.configure_optimizers(torch.optim.Adamax, lr = 0.001, weight_decay = 1e-0
 
 ```
 
-Then the Trainer can be trained and tested, and the model can be saved:
+Then the Trainer can be trained and tested; the best model in terms of validation loss is saved by default, and the user can modify so or indicate where to save it using the `train()` method parameters `save_model` and `filename`.
```
DaniBodor, line 293: same as in getstarted doc

tests/test_trainer.py
```diff
@@ -136,7 +135,7 @@ def test_grid_regression(self):
             CnnRegression,
             dataset
         )
-        trainer.train(nepoch=1, batch_size=2, save_best_model=None)
+        trainer.train(nepoch=1, batch_size=2, best_model=False, save_model=False)
```
DaniBodor, line 139: throughout this file, whenever `save_model=False`, you don't need to specify a `best_model` argument.

DaniBodor (approved):
Looks good. Thanks for catching my mistakes on the early stopper!
All my comments are suggestions to resolve as you think is best.

deeprankcore/trainer.py
gcroci2: I think you need it in any case. It determines which model is going to be used for later testing (last or best), even if the model itself is not saved in a file. 

I can remove save_model and make filename optional though.

tests/test_trainer.py
gcroci2: Same as above, it does change which model is going to be used for later testing. Even if it's true that in the sake of this test, it doesn't matter because we're not testing last vs best. I placed filename=None in the tests where we're not interested in the saved model.

Issue #429, opened 2023-05-22
problem with get_pssm when preprocessing variants


rgayatri:
When check_pssm is ignored, I face another error with get_pssm 
```

Traceback (most recent call last):
  File "/home/gayatrir/miniconda3/lib/python3.10/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/home/gayatrir/miniconda3/lib/python3.10/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/mnt/home2/gayatrir/deeprank-core/deeprankcore/query.py", line 244, in _process_one_query
    graph = query.build(feature_modules)
  File "/mnt/home2/gayatrir/deeprank-core/deeprankcore/query.py", line 609, in build
    feature_module.add_features(self._pdb_path, graph, variant)
  File "/mnt/home2/gayatrir/deeprank-core/deeprankcore/features/conservation.py", line 29, in add_features
    pssm_row = residue.get_pssm()
  File "/mnt/home2/gayatrir/deeprank-core/deeprankcore/molstruct/residue.py", line 54, in get_pssm
    raise FileNotFoundError(f'No pssm file found for Chain {self._chain}.')
FileNotFoundError: No pssm file found for Chain pdb2g98 B
```

My variant of interest is in chain A, thus the pssm_path I provide is for chain A. 
Code to reproduce.

```
from deeprankcore.query import QueryCollection, SingleResidueVariantAtomicQuery
from deeprankcore.features import components, conservation, contact, surfacearea
from deeprankcore.utils.grid import GridSettings, MapMethod
from deeprankcore.domain.aminoacidlist import (arginine, cysteine)
from deeprankcore.domain import targetstorage as targets

queries.add(SingleResidueVariantAtomicQuery(
    pdb_path = "/home/gayatrir/DATA/pdb/pdb2g98.ent",
    chain_id = "A",
    residue_number = 14,
    insertion_code = None,
    wildtype_amino_acid = arginine,
    variant_amino_acid = cysteine,
    pssm_paths = {"A": "/mnt/csb/DeepRank-Mut-DATA/pssm_update/2g98.A.pdb.pssm"},
    targets={targets.BINARY: 1},
    radius= 10.0,
    distance_cutoff= 4.5,
    ))

feature_modules = [components, conservation, contact, surfacearea]

hdf5_paths = queries.process(
    "/home/gayatrir/coretrain.hdf5",
[...]

PR #430, opened 2023-05-22
fix: check only 1 pssm for variant queries

### LINKED ISSUE #428: problem with check_pssm
I am trying to preprocess variant data using SingleResidueVariantAtomicQuery.
Here is the error I get

```
Graph/Query with ID atomic-graph:A:14:Arginine->Cysteine:pdb2g98 ran into an Exception (ValueError: Amino acids in PSSM files do not match pdb file for pdb2g98.ent.), and it has not been written to the hdf5 file. More details below:
Amino acids in PSSM files do not match pdb file for pdb2g98.ent.
Traceback (most recent call last):
  File "/mnt/home2/gayatrir/deeprank-core/deeprankcore/query.py", line 65, in _check_pssm
    if pdb_truth[residue] != pssm_data[residue]:
KeyError: 'B0001'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/mnt/home2/gayatrir/deeprank-core/deeprankcore/query.py", line 244, in _process_one_query
    graph = query.build(feature_modules)
  File "/mnt/home2/gayatrir/deeprank-core/deeprankcore/query.py", line 570, in build
    structure = self._load_structure(self._pdb_path, self._pssm_paths, include_hydrogens, load_pssms)
  File "/mnt/home2/gayatrir/deeprank-core/deeprankcore/query.py", line 129, in _load_structure
    _check_pssm(pdb_path, pssm_paths)
  File "/mnt/home2/gayatrir/deeprank-core/deeprankcore/query.py", line 68, in _check_pssm
    raise ValueError(error_message) #pylint: disable = raise-missing-from
ValueError: Amino acids in PSSM files do not match pdb file for pdb2g98.ent.
```

My task does not involve inclusion of multiple chains. 
Also, I note that residues in pdb and pssm are identical. 

Code to reproduce:
```
from deeprankcore.query import QueryCollection, SingleResidueVariantAtomicQuery
from deeprankcore.features import components, conservation, contact, surfacearea
from deeprankcore.utils.grid import GridSettings, MapMethod
from deeprankcore.domain.aminoacidlist import (arginine, cysteine)
from deeprankcore.domain import targetstorage as targets

queries.add(SingleResidueVariantAtomicQuery(
    pdb_path = "/home/gayatrir/DATA/pdb/pdb2g98.ent",
    chain_id = "A",
    residue_number = 14,
    insertion_code = None,
    wildtype_amino_acid = arginine,
    variant_amino_acid = cysteine,
    pssm_paths = {"A": "/mnt/csb/DeepRank-Mut-DATA/pssm_update/2g98.A.pdb.pssm"},
    targets={targets.BINARY: 1},
    radius= 10.0,
    distance_cutoff= 4.5,
    ))


```

DaniBodor:
- #428 fixed by only checking for residues on chains for which a pssm file was provided.
- #429 will not be fixed, because we want users to provide pssm files for all residues that are included in the graph.
- Created an option to suppress the error thrown by `_check_pssm` and give a warning instead.
    - does this require additional documentation, beyond the docstrings?
    - created tests for error suppression
- Create tests for variant query where there is a second chain for which no pssm file is provided:
    - throw error if second chain is inside `radius`
    - no error if second chain is NOT inside `radius`
- Refactor existing tests relating to _check_pssm
- Allow for `_check_pssm` to throw more verbose errors
     - this is only accessible from source code, which I think is ok because it's more of a troubleshooting/debugging mode thing than something that we expect the user to want to touch, right?

closes #428, #429


rgayatri:
I tried with running preprocessing with this branch. It seems it still tries to check for all chains in PDB? 
Unfortunately some PDBs have errors in chain names, so there are inconsistencies between no. of PDB chains and no. of PSSMs. 
I'll do checks again tomorrow

rgayatri:
The changes work. This pull request can be closed. 
However, there should be a provision to include other pssm chains in case the radius includes multiple chains. can be dealt in #429

DaniBodor:
> The changes work. This pull request can be closed. However, there should be a provision to include other pssm chains in case the radius includes multiple chains. can be dealt in #429

It works because it now fills the pssm feature with 0s. So this PR actually deals with both #428 and #429, as they are closely related. In retrospect, maybe that wasn't ideal, because now things are getting a bit confused.
I will await decision on #429 to know whether we actually want to fill it with 0s, or want to force pssm files to be present for all chains and then either create separate PRs or resolve both issues here.

rgayatri:
I am making a small dataset with variants that have neighbours in multiple chains. 
I suppose that may be the only way to test consequence of 0s or non-0s in predictions (non-0s will be ideal for these cases)

rgayatri:
My test was for the general applicability. By omitting pssm files and adding 0s we would treat such variants as though they are solvent-exposed variants- which is not quite right, although I am not sure how many such cases are there. I have never tested such a thing with deeprank-mut, but it is expected and I do note that solvent-exposed variants are treated differently in network.

I would prefer to include all chains within the radius, so may be user can definitely include all relevant chains. Rest can be 0

DaniBodor:
so do I understand correctly that all residues from chains within the radius should be included in the network and that for those chains a pssm file should be required?

DaniBodor:
@rgayatri , can you please test on the current branch if it still works as intended for you?

Note that `Query` objects now have an optional argument `suppress_pssm_errors`

deeprankcore/query.py
```diff
@@ -645,13 +683,14 @@ def _load_ppi_atoms(pdb_path: str,
     return contact_atoms
 
 
-def _load_ppi_pssms(pssm_paths: Union[Dict[str, str], None],
-                    chain_id1: str, chain_id2: str,
+def _load_ppi_pssms(pssm_paths: Optional[Dict[str, str]],
+                    chains: List[str],
```
gcroci2, line 687: Right now we don't support more than 2 chains for PPI, so I think it's better to have chain_id1 and chain_id2 to make it more explicit

gcroci2 (approved):
Very nice edits, and great job with the tests!
I have only a minor comment about an internal function in query.py.

For `_check_pssm`, I think that the docstring so far is enough.

About the verbosity, I do agree that it's okay if it's accessible internally only.

deeprankcore/query.py
DaniBodor: I made the change mainly because the linter was complaining, but also think it's nicer because a few lines below we anyway loop through the two chains turn them into a list there. 
I think because it's an internal function, it doesn't matter too much how explicit it is now. And your comment makes me realize that if in the future we would want to allow more than 2 chains in a PPI, then this is one less issue to think about :)

DaniBodor:
> I think we're missing a test that uses `suppress_pssm_errors=True`. Indeed I was wondering, what happens if the errors are suppressed and pssm files are missing or incorrect? What will conservation-related features contain?

lines 390 and 413 of test_query.py ;)

Good question what the hdf5 file will then contain. I will look into it.

gcroci2:
> > I think we're missing a test that uses `suppress_pssm_errors=True`. Indeed I was wondering, what happens if the errors are suppressed and pssm files are missing or incorrect? What will conservation-related features contain?
> 
> lines 390 and 413 of test_query.py ;-)
> 
> Good question what the hdf5 file will then contain. I will look into it.

Oh right! What if we explicitly set `suppress_pssm_errors` in the query initialization (maybe redefining the query in the test)? This way is clearer how such a parameter can be used, instead of setting the internal `_suppress`. Since we don't have explicit examples about `suppress_pssm_errors`, can be good to have it in the tests

DaniBodor:
> > > I think we're missing a test that uses `suppress_pssm_errors=True`. Indeed I was wondering, what happens if the errors are suppressed and pssm files are missing or incorrect? What will conservation-related features contain?
> > 
> > 
> > lines 390 and 413 of test_query.py ;-)
> > Good question what the hdf5 file will then contain. I will look into it.
> 
> Oh right! What if we explicitly set `suppress_pssm_errors` in the query initialization (maybe redefining the query in the tests)? This way is clearer how such a parameter can be used, instead of setting the internal `_suppress`. Since we don't have explicit examples about `suppress_pssm_errors`, can be good to have it in the tests

I did it this way to make sure that the test uses the same instance of the object that failed if the error suppression was not turned on (a few lines above). This way, if someone modifies the Query object in the test, we are sure that both tests run the same instance.
Do you think I should add an additional test where it is set upon initialization?

gcroci2:
> > > I think we're missing a test that uses `suppress_pssm_errors=True`. Indeed I was wondering, what happens if the errors are suppressed and pssm files are missing or incorrect? What will conservation-related features contain?
> > 
> > 
> > lines 390 and 413 of test_query.py ;-)
> > Good question what the hdf5 file will then contain. I will look into it.
> 
> OK, I looked into it:
> 
> * for incorrect entries in the pssm: it just used the "wrong" data. So the pssm data that is associated to the position number of that residue, even though it is the incorrect residue (@rgayatri is this the behavior that you want?)
> * for missing entries in the pssm:
>   
>   * if the entry is not part of the graph, it doesn't complain
>   * if the entry is a part of the graph, then it fails at a later step (so `check_pssm` passes, but then an error is thrown I think when it tries to write data into that key).
> 
> The latter point made me realize that it still checks all residues from the chain, irrespective of whether they are a part of the graph. I will rewrite so that it only checks those residues that are part of the graph.
> 
> @rgayatri , what behavior do you expect if a residue is missing from a pssm file? All 0s?
> 
> At the end of the day, all these tests make me feel that we just don't want people to use bad pssm files. I don't like the concept of being able to suppress the error, because now we need to hack ad hoc solutions to deal with bad data.

I agree, ideally, a user shouldn't be able to suppress the error indeed and should take care of providing the correct files.

DaniBodor:
I believe that at this point the current PR is functional. If fed with accurate data, everything works as expected. Otherwise errors can be suppressed if absolutely necessary, but it can still fail if this touches data points that are part of the graph. This is okay behavior, as we do not want users to train with unexpected/unexplainable results.
As such, I will be merging and closing this PR. If there are any remaining problems to be solved about this, we can open new issues about them.

PR #433, opened 2023-05-26
ci: automatically check markdown links

DaniBodor:
added a CI workflow to check that all links in md files are active


.mlc-config.json
```diff
@@ -0,0 +1,37 @@
+{
```
gcroci2, line 1: Is this file needed?

gcroci2 (approved):

gcroci2 (approved):

.mlc-config.json
DaniBodor: yes, but only for 1 link that is incorrectly flagged as broken. I minimized the file and moved it out of the root.

PR #443, opened 2023-06-02
docs: update README.md

gcroci2:
The current PyPi release does not work with the documentation we have, because the release is outdated. I'll do a new release very soon but in the meantime we need to add that to the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ Once the dependencies installed, you can install the latest release of deeprankc
 pip install deeprankcore
 ```
 
-Please note that the PyPi release is currently outdated and we will do a new release very soon. In the meantime, you can get all the new developments by cloning the repo and installing the code with
+You can get all the new developments by cloning the repo and installing the code with
 
 ```
 git clone https://github.com/DeepRank/deeprank-core
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ DeeprankCore extensive documentation can be found [here](https://deeprankcore.rt
 
 ### Dependencies
 
-Before installing deeprankcore you need to install:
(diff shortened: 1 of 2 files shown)


DaniBodor:
what do you mean by it doesn't work?

gcroci2:
> what do you mean by it doesn't work?

The docs are updated with the code, but the release is old so if you pip install deeprankcore using the released version (1.0.0) and then try to follow the docs for using the package, you'll get errors and issues.

README.md
```diff
@@ -80,7 +80,7 @@ Once the dependencies installed, you can install the latest release of deeprankc
 pip install deeprankcore
 ```
 
-You can get all the new developments by cloning the repo and installing the code with
+Please note that the PyPi release is currently outdated and we will do a new release very soon. In the meantime, you can get all the new developments by cloning the repo and installing the code with
```
DaniBodor, line 83: it's still unclear to me what the added value of just this statement is. If people can use the package, I don't think we need to say this. If they can't use the package right now, we shouldn't tell them to install.

If we do need to say something about it, I think it needs a bit more explanation. As a non-expert user, if I see this statement, I would have no idea what I should do now.

DaniBodor (changes requested):
Maybe also remove reduce from install instructions already? Even though the other thing isn't ready to be announced yet, reduce is not useful and only gives bugs anyway and is the most tedious part of the install instructions.

gcroci2:
> Maybe also remove reduce from install instructions already? Even though the other thing isn't ready to be announced yet, reduce is not useful and only gives bugs anyway and is the most tedious part of the install instructions.

Done, I've also removed reduce from the CI, since we're not testing add_hydrogens it shouldn't break anything.

DaniBodor (approved):
awesome!

PR #434, opened 2023-05-30
docs: add tutorials for PPIs

gcroci2:
With this PR, I add two tutorial notebooks for PPI data;
-  `data_generation_ppi.ipynb`
   - residue level data generation
   - atomic level data generation
-  `training_ppi.ipynb`
   - classification for GNNs using `GraphDatasets`
   - classification for CNNs using `GridDatasets`
  - references to regression across the notebook

PR #446, opened 2023-06-13
refactor: inherit information from training set for valid/test sets

joyceljy:
### ```dataset.py``` & ```test_dataset.py``` ###
1. Add a function named ```_check_inherited_param```. For validation and test datasets(while ```train``` is set as False), ```_check_inherited_param``` is performed and the following information will be inherited from the training set and be given a warning message:
```target```, ```task```, ```node_features``` , ```edge_features```, ```features_dict```, ```target_transform``` and ```classes```.
2. Modify the order of parameters in ```GraphDataset``` &  ```GridDataset```object to the followings:
hdf5_path, subset, train, dataset_train, node_features, edge_features, features_transform, clustering_method, target, target_transform, target_filter, task, classes, tqdm, root, check_integrity
3. Change the unit tests  ```GraphDataset``` &  ```GridDataset``` parameters order, and remove the parameters that will be inherited from the training set.
4. Add unit tests ```test_inherit_info_training_griddataset``` & ```test_inherit_info_training_griddataset``` for checking whether the above parameters is inherited.
5. Add a unit test ```test_incompatible_dataset_train_type``` for checking the compatibility of assigning different types of data objects for training, validation, and testing dataset.

### ```trainer.py``` & ```test_trainer.py``` ###
1. Modify function ```_check_dataset_equivalence```, remove the for loop of checking parameters of dataset (because these information should have been inherited directly).
2. Add function ```_check_dataset_value``` for checking up the settings for valid/test dataset.
3. Modify unit test ```test_dataset_equivalence_no_pretrained``` & ```test_dataset_equivalence_pretrained``` to match the new trainer.py.
4. Remove unit test ```test_grid_graph_incompatible```, this test is related to dataset and should be done in test_dataset.py.

### Tutorials/ReadME/other files ###
Change the description of  ```GraphDataset``` &  ```GridDataset``` objects' parameters order, and remove the parameters that will be inherited from the training set.

### CODE REVIEW

File: deeprankcore/dataset.py [hunk 1]
```diff
@@ -658,17 +658,40 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
...
             if not isinstance(dataset_train, GraphDataset):
                 raise TypeError("Please provide a valid training GraphDataset.")
-
-            if dataset_train.features_transform is not None:
-                self.features_transform = dataset_train.features_transform
-            else:
-                self.features_transform = None
+            
+            #set features_transform as the same in dataset_train
+            self.features_transform = dataset_train.features_transform
```
gcroci2, line 663: The if statement should stay, since it may be that `dataset_train.features_transform` is None. In that case, we want `self.features_transform` to be None as well.

deeprankcore/dataset.py
```diff
@@ -658,17 +658,43 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
...
             else:
                 self.features_transform = None
             _log.warning("features_transform parameter, if set, will be ignored." + "\n" +
-                        "features_transform will remain the same as the one used in training phase.")
+                        "features_transform will remain the same as the one used in training phase: %s.", self.features_transform)
+
+            #set task as the same in dataset_train
+            self.task = dataset_train.task
+            _log.warning("task parameter, if set, will be ignored." + "\n" +
+                        "task will remain the same as the one used in training phase: %s.", self.task)
```
DaniBodor, line 673: I feel that we only need to add these warnings (or maybe even throw errors?) in case a user set the parameters manually and they differ from the inherited one.
If that doesn't happen, I don't see any reason why this needs to go into the log or warnings or anywhere. It is the default expected behavior of the software.
One way to do this is to loop through the different inherited settings and then check for each whether it matches the user set one and deal accordingly.

Also, i personally very much prefer f-strings over %-formatting (plus I think it's the [recommended way](https://realpython.com/python-f-strings/) now).

DaniBodor (changes requested):
It would be nice to update the docstring to reflect that the parameters should only be set for training dataset. For that matter, I think that the `dataset_train` argument should precede the ones that are inherited and that is where the docstring can then state which parameters should be ignored when set to False. 

Also I'd consider making `dataset_train` a required parameter, but not sure about that.

DaniBodor:
PS: I changed the type of this PR, because it's not actually a bug fix, just a nicer way of dealing with things :)

deeprankcore/dataset.py
joyceljy: I don't think I understand fully what you mentioned about `dataset_train`.
Do you mean to move its order in the definition before the parameters that needed to be inherited?
Besides, I modified the docstring to state the parameters that will be ignored and inherited.

deeprankcore/dataset.py
```diff
@@ -581,24 +581,36 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
                 Defaults to None.
             subset (Optional[List[str]], optional): List of keys from .HDF5 file to include. 
                 Defaults to None (meaning include all).
+                
+            dataset_train (class:`GraphDataset`, optional): If train is True, assign here the training set,
+                the following parameters target, task, node_features, edge_features and features_transform will be set
+                as the same value in dataset_train. 
+
+                Defaults to None.
+
```
DaniBodor, line 590: I think I didnt explain well in my previous comment. 

Right now, it's still a bit confusing, especially because the `train` parameter is set later.

I think it makes sense to first set `train` to True or False and in the description of that argument state: if `True` then set these arguments and if `False` set the other arguments. Then in the description of the other arguments state that they will be ignored if the `train` setting is False/True.

joyceljy: Thanks for the explanation:)
True that the order of the definition is quite confusing, 
I adjust the order of the parameters and the description in docstring a bit.
Now it should be more readable.

DaniBodor (approved):
Yes, much clearer. Thanks.
Now only the `train` argument in the actual initialization should match the location in the docstring (so line 574 also needs to move above line 562). Then I think it's all done 👍

deeprankcore/dataset.py
```diff
@@ -581,24 +581,36 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
                 Defaults to None.
             subset (Optional[List[str]], optional): List of keys from .HDF5 file to include. 
                 Defaults to None (meaning include all).
+            train (bool, optional): Boolean flag to determine if the instance represents the training set. If False, a dataset_train of the same class must
+                be provided as well. The latter will be used to scale the validation/testing set according to its features values.
+                Defaults to True.    
+            dataset_train (class:`GraphDataset`, optional): If train is True, assign here the training set,
+                the following parameters target, task, node_features, edge_features and features_transform will be set
```
gcroci2, line 588:
```suggestion
                If `train` is True and `dataset_train` is assigned, the parameters target, task, node_features, edge_features and features_transform will be set
```

deeprankcore/dataset.py
```diff
@@ -581,24 +581,36 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
...
+                as the same value in dataset_train. 
+                Defaults to None.
             target (Optional[str], optional): Default options are irmsd, lrmsd, fnat, binary, capri_class, dockq, and BA.
                 It can also be a custom-defined target given to the Query class as input (see: `deeprankcore.query`); 
                 in this case, the task parameter needs to be explicitly specified as well.
                 Only numerical target variables are supported, not categorical. 
                 If the latter is your case, please convert the categorical classes into
-                numerical class indices before defining the :class:`GraphDataset` instance. Defaults to None.
+                numerical class indices before defining the :class:`GraphDataset` instance.
+                Value will be ignored and inherited from the training set if train set as False.
```
gcroci2, line 597:
```suggestion
                Value will be ignored and inherited from `dataset_train` if train set as False. 
```

deeprankcore/dataset.py
```diff
@@ -618,23 +630,15 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
...
-                Defaults to None.
-                
+                Defaults to True.    
             features_transform (Optional[dict], optional): Dictionary to indicate the transformations to apply to each feature in the dictionary, being the
                 transformations lambda functions and/or standardization.
                 Example: `features_transform = {'bsa': {'transform': lambda t:np.log(t+1),' standardize': True}}` for the feature `bsa`.
                 An `all` key can be set in the dictionary for indicating to apply the same `standardize` and `transform` to all the features.
                 Example: `features_transform = {'all': {'transform': lambda t:np.log(t+1), 'standardize': True}}`.
                 If both `all` and feature name/s are present, the latter have the priority over what indicated in `all`.
+                Value will be ignored and inherited from the training set if train set as False.
```
gcroci2, line 641:
```suggestion
                Value will be ignored and inherited from `dataset_train` if `train` is set as False and `dataset_train` is assigned. 
```

deeprankcore/dataset.py
```diff
@@ -581,24 +581,36 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
...
             task (Optional[str], optional): 'regress' for regression or 'classif' for classification. Required if target not in
                 ['irmsd', 'lrmsd', 'fnat', 'binary', 'capri_class', 'dockq', or 'BA'], otherwise this setting is ignored.
                 Automatically set to 'classif' if the target is 'binary' or 'capri_classes'.
                 Automatically set to 'regress' if the target is 'irmsd', 'lrmsd', 'fnat', 'dockq' or 'BA'.
+                Value will be ignored and inherited from the training set if train set as False. 
                 Defaults to None.
             node_features (Optional[Union[List[str], str], optional): Consider all pre-computed node features ("all") or
                 some defined node features (provide a list, example: ["res_type", "polarity", "bsa"]).
                 The complete list can be found in `deeprankcore.domain.nodestorage`.
+                Value will be ignored and inherited from the training set if train set as False.
```
gcroci2, line 608:
```suggestion
                Value will be ignored and inherited from `dataset_train` if `train` is set as False and `dataset_train` is assigned. 
```

deeprankcore/dataset.py
```diff
@@ -658,17 +662,49 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
         if not train:
             if not isinstance(dataset_train, GraphDataset):
                 raise TypeError("Please provide a valid training GraphDataset.")
-
+            
+            #set features_transform as the same in dataset_train
             if dataset_train.features_transform is not None:
+                if self.features_transform != dataset_train.features_transform:
+                    _log.warning("features_transform parameter, if set, will be ignored." + "\n" +
```
gcroci2, line 669:
```suggestion
                    _log.warning(f"Overwriting features_transform parameter with the one used in the training phase: {self.features_transform}.")
```

deeprankcore/dataset.py
```diff
@@ -658,17 +662,49 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
...
+                        f"features_transform will remain the same as the one used in training phase: {self.features_transform}.")
```
gcroci2, line 670:
```suggestion

```

tests/test_dataset.py
```diff
@@ -857,6 +857,39 @@ def test_features_transform_logic_graphdataset(self):
         assert dataset_train.means == dataset_test.means
         assert dataset_train.devs == dataset_test.devs
 
+    def test_inherit_info_training(self):
```
gcroci2, line 860: I'd test two cases: the first in which I assign node_features, edge_features, features_transform, target and task in the training dataset only, and I verify that if I define a test dataset flagging train as False and passing in dataset_train the attributed are the same; then a second case as you did below, in which I define a training set and then I re-assign parameters in the test set, and I verify that they are still the same as in the training set.

gcroci2 (changes requested):
Nice additions :) 
I added minor notes and a few comments before finishing this PR (almost there!!):
- I'd suggest to create an internal function (e.g., `_check_inherited_param`) which checks for all the parameters which should be inherited if train is False and dataset_train is given (e.g., a list like param_to_check = [node_features, ...] to be compared with the training parameters), and in the case in which they are not the same it overwrites the parameter and prints out the warning. This way we'll avoid to repeat multiple times very similar code and this check will be nicely packed in only one place. 
- I'd change the parameters order in GraphDataset (init method) in the following: hdf5_path, subset, train, dataset_train, node_features, edge_features, features_transform, clustering_method, target,  target_transform, target_filter, task, classes, tqdm, root, check_integrity
- I think that `target_transform` and `classes` should be inherited from the train set as well 
  - Do you agree? @DaniBodor)
- Since now in train=False cases we're making the dataset inherits from the dataset_train, I'd change the documentation by removing node_features, edge_features, ... assegnation from the validation/test sets instantiations (README.md, docs/getstarted.md), and from the tests that are not specifically testing the inheritance. 
- I'd also change the order of the parameters in the above mentioned documentation according to the order I mentioned in the second bullet point, just to be coherent with the code base (it's not essential for letting the code work because it works anyway since we're explicitly assigning variables, but it does help the user to see always the same order)

DaniBodor:
Also, make sure the README and docs/getstarted documents get updated with these changes (and the tutorial if #434 is closed before this).

gcroci2:
> Also, make sure the README and docs/getstarted documents get updated with these changes (and the tutorial if #434 is closed before this).

I am closing #434 now. So keep in mind that tutorials as well need to be updated with the order of the new parameters, and in general remember to remove the not needed parameters from validation/test sets @joyceljy

deeprankcore/dataset.py
```diff
@@ -658,17 +663,20 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
         if not train:
             if not isinstance(dataset_train, GraphDataset):
                 raise TypeError("Please provide a valid training GraphDataset.")
-
+            
+            #check inherited parameter with the ones in the training set
+            inherited_param = ["target", "task", "node_features", "edge_features", "features_dict", "classes", "target_transform", "features_transform"]
+            self._check_inherited_param(inherited_param, dataset_train)
+            
+            #set features_transform as the same in dataset_train
```
gcroci2, line 671: Line 672 and below now are not needed anymore (that check is in `_check_inherited_param`)

deeprankcore/dataset.py
```diff
@@ -685,7 +693,30 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
                 dataset_train._compute_mean_std()
             self.means = dataset_train.means
             self.devs = dataset_train.devs
-
+            
+    def _check_inherited_param(
+        self,
+        inherited_param: List[str] = None,
```
gcroci2, line 699: inherited_param and dataset_train shouldn't be optional, since the check function is called only when you actually have both inherited_param and dataset_train assigned

deeprankcore/dataset.py
```diff
@@ -685,7 +693,30 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
...
+        dataset_train: GraphDataset = None,
+    ):
+        """Check if the parameters from validation and testing are same as the training set.
+        
+        Args:
+        inherited_param (List[str], optional): List of parameters that need to be check for inheritance.
```
gcroci2, line 705:
```suggestion
        inherited_param (List[str], optional): List of parameters that need to be checked for inheritance.
```

deeprankcore/dataset.py
```diff
@@ -685,7 +693,30 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
...
+            Defaults to None.
+        dataset_train (class:`GraphDataset`, optional): The parameters in `inherited_param` will be inherited from `dataset_train`.
+            Defaults to None.
+        """
+        
+        self_vars = vars(self)
```
gcroci2, line 711: This could be further simplified (line 711 - 718):

```python
        for param in inherited_param:
            train_param = getattr(dataset_train, param)
            if getattr(self, param) != train_param:
                setattr(self, param, train_param)
                _log.warning(f"Overwriting {param} parameter with the one " +
                            f"used in the training phase: {train_param}.")   
```

tests/test_dataset.py
```diff
@@ -846,17 +846,72 @@ def test_features_transform_logic_graphdataset(self):
...
+        
+        assert dataset_train.node_features == dataset_valid.node_features
+        assert dataset_train.edge_features == dataset_valid.edge_features
+        assert dataset_train.features_dict == dataset_valid.features_dict
+        assert dataset_train.features_transform == dataset_valid.features_transform
+        assert dataset_train.target == dataset_valid.target
+        assert dataset_train.target_transform == dataset_valid.target_transform
+        assert dataset_train.task == dataset_valid.task
+        
+        dataset_test = GraphDataset(
```
gcroci2, line 892:
```suggestion
        dataset_valid = GraphDataset(
```
Let's stick to the same name

tests/test_trainer.py
```diff
@@ -49,20 +49,22 @@ def _model_base_test( # pylint: disable=too-many-arguments, too-many-locals
...
-        target=target,
-        task = task,
         clustering_method=clustering_method,
-        target_transform = target_transform)
+        target=target,
+        target_transform = target_transform,
+        task = task
+        )
 
     if val_hdf5_path is not None:
```
gcroci2, line 58: Same here as in test_integration.py; we need to specify `train` and `dataset_train`, and we can avoid specifying node_features, ... in the validation and test sets.

deeprankcore/dataset.py
```diff
@@ -581,25 +581,31 @@ def __init__( # pylint: disable=too-many-arguments, too-many-locals
...
-                in this case, the task parameter needs to be explicitly specified as well.
-                Only numerical target variables are supported, not categorical. 
-                If the latter is your case, please convert the categorical classes into
-                numerical class indices before defining the :class:`GraphDataset` instance. Defaults to None.
-            task (Optional[str], optional): 'regress' for regression or 'classif' for classification. Required if target not in
-                ['irmsd', 'lrmsd', 'fnat', 'binary', 'capri_class', 'dockq', or 'BA'], otherwise this setting is ignored.
-                Automatically set to 'classif' if the target is 'binary' or 'capri_classes'.
-                Automatically set to 'regress' if the target is 'irmsd', 'lrmsd', 'fnat', 'dockq' or 'BA'.
+            train (bool, optional): Boolean flag to determine if the instance represents the training set. If False, a dataset_train of the same class must
+                be provided as well. The latter will be used to scale the validation/testing set according to its features values.
```
gcroci2, line 585:
```suggestion
                be provided as well. The latter will be used to scale the validation/testing set according to its features values and to match the datasets' parameters.
```

tests/test_trainer.py
```diff
@@ -49,20 +49,22 @@ def _model_base_test( # pylint: disable=too-many-arguments, too-many-locals
...
         dataset_val = GraphDataset(
             hdf5_path=val_hdf5_path,
             node_features=node_features,
             edge_features=edge_features,
-            target=target,
-            task = task,
             clustering_method=clustering_method,
-            target_transform = target_transform)
+            target=target,
```
gcroci2, line 64: Same as above, in validation and testing case we need to match the code with the train and dataset_train logic

gcroci2 (changes requested):
Great improvements! This PR turned out to be much bigger than expected but it's improving the training logic quite a lot :-)
I left few comments, and I realized that we still need to address a couple of issues:
- We need to move the internal method that checks for parameters into the parent class `DeeprankDataset`, where indeed also the other checkings are defined. The reason for it is that we want to do such checks both in GraphDataset and GridDataset. We also need to add `train` and `dataset_train` parameters in GridDataset as well, as well as the parameters' checking. GridDataset will only be missing `features_transform` (no standardization and no transformation), but we're fine with that (for now). I do think that the training logic should be in both dataset types. 
  - Since GridDataset has only `features`, and not `node_features` and `edge_features`, we will need an if statement that checks for the right thing depending on the dataset class. 
- Since we're now enforcing valid/test sets to have the same parameters as the training set, I would make it the standard way of defining such sets, inserting a check in the Trainer as well: if the user defines dataset_train, dataset_valid (and eventually dataset_test) in the Trainer, then dataset_valid (and dataset_test) need to have the `train` flag set to False, which can become an attribute of the GraphDataset class to easily access it afterwards. Otherwise, the Trainer should give an error about the validation/test sets. This way we can basically remove the for loop in `_check_dataset_equivalence` (in trainer.py). What do you think? @DaniBodor @joyceljy

DaniBodor:
> I'm a little bit confused on what `_check_dataset_equivalence` function should check. From my understanding, I think this `_check_dataset_equivalence` should check the following things:
> 
> 1. If train_dataset is None, a test_dataset should be exsists (This happens in the case `pretrain_model` is None), and make sure test_dataset is a type of Graph/Grid dataset.
> 
> 2. If valid_dataset/test_dataset exsits, make sure valid_dataset/test_dataset should have same dataset as train_dataset.
> 
> 3. If valid_dataset/test_dataset exsits, check `train` parameter in valid_dataset/test_dataset is set as False.
> 
> 4. If valid_dataset/test_dataset exsits, check `train_dataset` parameter is set and is equivalent to train_dataset.
>        Please correct me if I misunderstood something

1. correct, so no change there
2. check that they are of the same type (not sure if that's what you meant)
3. correct, and also check that `train` in `dataset_train` is set to `True`
4. yes, but I think this is the same as 2

DaniBodor (changes requested):
Indeed turning into a much bigger PR than expected.
Great job handling it, @joyceljy !

Apart from inline comments, shouldn't `def len(self)` in dataset.py be `def __len__(self)`

joyceljy:
> > I'm a little bit confused on what `_check_dataset_equivalence` function should check. From my understanding, I think this `_check_dataset_equivalence` should check the following things:
> > 
> > 1. If train_dataset is None, a test_dataset should be exsists (This happens in the case `pretrain_model` is None), and make sure test_dataset is a type of Graph/Grid dataset.
> > 2. If valid_dataset/test_dataset exsits, make sure valid_dataset/test_dataset should have same dataset as train_dataset.
> > 3. If valid_dataset/test_dataset exsits, check `train` parameter in valid_dataset/test_dataset is set as False.
> > 4. If valid_dataset/test_dataset exsits, check `train_dataset` parameter is set and is equivalent to train_dataset.
> >    Please correct me if I misunderstood something
> 
> 1. correct, so no change there
> 2. check that they are of the same type (not sure if that's what you meant)
> 3. correct, and also check that `train` in `dataset_train` is set to `True`
> 4. yes, but I think this is the same as 2

Thanks Dani!
2. is to check the dataset type(whether it's graphdataset or griddataset), other type will raise an error.
4. is to check whether the ```dataset_train``` parameter is set and equivalent to the train dataset.

deeprankcore/dataset.py
```diff
@@ -378,6 +380,29 @@ def _compute_mean_std(self):
...
+        """
+        
+        self_vars = vars(self)
+        dataset_train_vars = vars(dataset_train)
+
+        for param in inherited_param:
+            if (self_vars[param] != dataset_train_vars[param]):
+                setattr(self, param, dataset_train_vars[param])
+                _log.warning(f"Overwriting {param} parameter with the one " +
+                            f"used in the training phase: {self_vars[param]}.")
```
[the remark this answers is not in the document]
joyceljy: I will expand the warning message to make it more expressive:)

deeprankcore/dataset.py
```diff
@@ -390,44 +415,63 @@ def __init__( # pylint: disable=too-many-arguments
...
+        dataset_train: GridDataset = None,
+        features: Optional[Union[List[str], str]] = "all",
         target: Optional[str] = None,
+        target_transform: Optional[bool] = False,
+        target_filter: Optional[Dict[str, str]] = None,
         task: Optional[str] = None,
-        features: Optional[Union[List[str], str]] = "all",
         classes: Optional[Union[List[str], List[int], List[float]]] = None,
         tqdm: Optional[bool] = True,
         root: Optional[str] = "./",
```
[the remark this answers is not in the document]
joyceljy: I'm not pretty sure about the ```tqdm``` & ```target_transform``` argument.
Do you have suggestions for that? @gcroci2

gcroci2:
> I'm a little bit confused on what `_check_dataset_equivalence` function should check. From my understanding, I think this `_check_dataset_equivalence` should check the following things:
> 
> 1. If train_dataset is None, a test_dataset should be exsists (This happens in the case `pretrain_model` is None), and make sure test_dataset is a type of Graph/Grid dataset.

In such case (train_dataset None and test/valid dataset not None), `pretrain_model` can't be None; there is no training data, so the only way for make it work is to have al already trained model to be tested on some testset. @joyceljy

deeprankcore/dataset.py
```diff
@@ -390,44 +415,63 @@ def __init__( # pylint: disable=too-many-arguments
         self,
         hdf5_path: Union[str, list],
         subset: Optional[List[str]] = None,
+        train: bool = True,
+        dataset_train: GridDataset = None,
+        features: Optional[Union[List[str], str]] = "all",
         target: Optional[str] = None,
+        target_transform: Optional[bool] = False,
```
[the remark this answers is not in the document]
gcroci2: Indeed, it should be just a bool, defaulted to False @joyceljy

gcroci2:
> Indeed turning into a much bigger PR than expected.
> Great job handling it, @joyceljy !

> Apart from inline comments, shouldn't def len(self) in dataset.py be def __len__(self)

This is also not in the scope of this PR, let's try to not make it even bigger and with more scopes than it already has ;) (even if it would be a minor change). 

About the __len__ method, DeeprankDataset inherits from `torch_geometric.data.dataset.Dataset` and overwrites the [Dataset's len method](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.data.Dataset.html#torch_geometric.data.Dataset.len). What is happening under the hoods is that len is calling the magic method __len__

@DaniBodor

deeprankcore/dataset.py
```diff
@@ -378,6 +380,29 @@ def _compute_mean_std(self):
         self.means = means
         self.devs = devs
 
+    def _check_inherited_param(
```
gcroci2, line 383: Please move this method after `_check_task_and_classes` (no functional changes)

deeprankcore/dataset.py
```diff
@@ -378,6 +380,29 @@ def _compute_mean_std(self):
...
+        self,
+        inherited_param: List[str],
+        dataset_train: Union[GraphDataset, GridDataset],
+    ):
+        """Check if the parameters from validation or testing are same as the training set.
```
gcroci2, line 388:
```suggestion
        """Check if the parameters for validation and/or testing are the same as in the training set.
```

deeprankcore/dataset.py
```diff
@@ -378,6 +380,29 @@ def _compute_mean_std(self):
...
+        
+        Args:
+        inherited_param (List[str]): List of parameters that need to be checked for inheritance.
+            Defaults to None.
```
gcroci2, line 392:
```suggestion

```

deeprankcore/dataset.py
```diff
@@ -378,6 +380,29 @@ def _compute_mean_std(self):
...
+        dataset_train (Union[class:`GraphDataset`, class:`GridDataset`]): The parameters in `inherited_param` will be inherited from `dataset_train`.
+            Defaults to None.
```
gcroci2, line 394:
```suggestion

```

deeprankcore/dataset.py
```diff
@@ -378,6 +380,29 @@ def _compute_mean_std(self):
...
 
+    def _check_inherited_param(
+        self,
+        inherited_param: List[str],
+        dataset_train: Union[GraphDataset, GridDataset],
+    ):
+        """Check if the parameters from validation or testing are same as the training set.
+        
+        Args:
+        inherited_param (List[str]): List of parameters that need to be checked for inheritance.
```
gcroci2, line 391:
```suggestion
        inherited_params (List[str]): List of parameters that need to be checked for inheritance.
```

Please change it also below

joyceljy:
I found out at last that the whether the type of ```dataset_train``` parameter is a Graph/GridDataset has already been checked in ```dataset.py``` file(See line 490 & 725). So the ```_check_dataset_equivalence``` will only check the following stuffs:
1. If train dataset is not set (this happens when a ```pretrained_model``` is provided), Trainer should have a test dataset assigned.
2. If train dataset is set, check whether the type of train dataset is a Graph/GridDataset.
3. Check ```train``` parameter in val/test dataset is set as False.
4. Check ```dataset_train``` parameter is equivalent to the train dataset which is passed to Trainer. 

Besides, I removed the test ```test_grid_graph_incompatible``` in ```test_trainer.py``` file since the checking of dataset type is actually done in dataset.py. Instead, a new test named ```test_incompatible_dataset_train_type``` is added to ```test_dataset.py```.

tests/test_trainer.py
```diff
@@ -148,27 +148,6 @@ def test_grid_classification(self):
...
-            dataset_train = dataset_train
-        )
-
-        with pytest.raises(TypeError):
-            Trainer(
-                CnnClassification,
-                dataset_train = dataset_train,
-                dataset_val = dataset_valid
-            )
-
```
DaniBodor: why did this test disappear? Or did it just move and I'm not finding it?

DaniBodor (commented):
Note that parameters `target_transform` and `tqdm` and `root` in both Dataset types are non `Optional`, but must be set (i.e., they cannot be None).

I realize it is just outside of the scope of this PR, but it is closely related and very easy to solve and would likely lead to merge conflicts if changed elsewhere, so I think it makes sense to change it here.

DaniBodor:
what a monster PR this has turned out to be :facepalm: 👹

deeprankcore/trainer.py
```diff
@@ -236,66 +236,45 @@ def _init_from_dataset(self, dataset: Union[GraphDataset, GridDataset]):
...
+                # Check dataset_train parameter in valid is equivalent to train which passed to Trainer.
+                if dataset_val.dataset_train != dataset_train:
+                    raise ValueError("valid dataset has different dataset_train parameter compared to the one given in Trainer.\n" +
+                                "Make sure to assign equivalent dataset_train in Trainer")
+                    
+            if dataset_test is not None:
+                # Check train parameter in test is set as False.
+                if dataset_test.train is not False:
+                    raise ValueError(f"""test dataset has train parameter {dataset_test.train}
+                                Make sure to set it as False""")
```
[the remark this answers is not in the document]
gcroci2: The issue would be about standardization, which is the reason why train and dataset_train have been created in the first place. In the beginning, they were meant for retaining and using means and stds from the training set in the case of testing/validation sets creation. I think that such distinction should happen when the sets are defined, which is when we instantiate the different GraphDataset objects. 

I still have to finish the new review of this PR, but what I think could be the least redundant solution can be the following: create sets and use train/dataset_train parameters in the DeeprankDataset children, and then in the Trainer only verify that what is supposed to be train, valid, test has the right settings to be so. Otherwise signal to the user that the datasets configurations are wrong

deeprankcore/trainer.py
```diff
@@ -236,66 +236,45 @@ def _init_from_dataset(self, dataset: Union[GraphDataset, GridDataset]):
...
+            
+            if dataset_val is not None:
+                # Check train parameter in valid is set as False.
+                if dataset_val.train is not False:
+                    raise ValueError(f"""valid dataset has train parameter {dataset_val.train}
+                                Make sure to set it as False""")
+                # Check dataset_train parameter in valid is equivalent to train which passed to Trainer.
+                if dataset_val.dataset_train != dataset_train:
+                    raise ValueError("valid dataset has different dataset_train parameter compared to the one given in Trainer.\n" +
+                                "Make sure to assign equivalent dataset_train in Trainer")
```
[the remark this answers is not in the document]
gcroci2: I still have to review it, but it shouldn't be assigned twice, the second time in the Trainer should be a check only.

deeprankcore/trainer.py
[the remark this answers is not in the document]
DaniBodor: I see the point, and think you are right. I also think that that is the way it is set up now, so then my comments can be disregarded.

gcroci2:
I left minor comments around, we will discuss this tomorrow but from my side I think it's fine as it is :)

joyceljy:
I also changed the description of this PR to give an overview of what this pr has solved!

docs/getstarted.md
```diff
@@ -173,25 +173,22 @@ target = "binary"
 # Creating GraphDataset objects
 dataset_train = GraphDataset(
     hdf5_path = hdf5_paths,
-    subset = train_ids, 
+    subset = train_ids,
```
gcroci2, line 176: Please add in this file as well (docs/getstarted.md) `train_ids = [<ids>]` and the other ids def as you did in the README :)

docs/getstarted.md
```diff
@@ -218,8 +215,8 @@ dataset_train = GraphDataset(
     subset = train_ids, 
     node_features = node_features,
     edge_features = edge_features,
-    target = target,
-    features_transform = features_transform
+    features_transform = features_transform,
```
gcroci2, line 218: Same as above, add train_ids = ... and the rest. Please do also for GridDataset

tests/test_dataset.py
```diff
@@ -118,19 +120,30 @@ def _compute_features_with_get(
                     features_dict[feat + f'_{ch}'] = arr
     return features_dict
 
+def _check_inherited_params(
```
gcroci2, line 123: No need or redefine it, you can import the method from dataset.py

gcroci2 (approved):
Great job :-) and finally over! Please address the two very minor comments that I left, then feel free to merge it and close the issue!

DaniBodor (approved):
I won't review the newest changes anymore and assume that @gcroci2 's final review suffices. So I will accept now, so that you have permission to merge.
Great job tackling this PR!!

One final comment I have, but super unimportant to solve (but very quick), is that I noticed that you (accidentally?) sometimes added some trailing whitespaces to some of the lines. I recently discovered a way to make VS code [automatically scrape trailing whitespaces](https://stackoverflow.com/a/53663494/648162) (in the File>Preferences>Settings search and tick "Trim trailing whitespace").
Also, you can auto replace all trailing whitespaces with a couple of clicks as described [here](https://stackoverflow.com/a/75537711/5170442)

PR #447, opened 2023-06-14
fix:  pdb files with underscore in the filename gives unexpected query ids

joyceljy:
Before the query ids will be seen as duplicates and be assigned a new id when the pdb files have the same base name (file name before the underscore character). 
For example:
```
pdb_paths = [
    str(PATH_TEST / "data/pdb/1ATN/1ATN_1w.pdb"),
    str(PATH_TEST / "data/pdb/1ATN/1ATN_2w.pdb"),
    str(PATH_TEST / "data/pdb/1ATN/1ATN_3w.pdb"),
    str(PATH_TEST / "data/pdb/1ATN/1ATN_4w.pdb")]
```
will give warnings as followings and rename the pdb files:
```
Query with ID residue-ppi:A-B:1ATN has already been added to the collection. Renaming it as residue-ppi:A-B:1ATN_2
Query with ID residue-ppi:A-B:1ATN has already been added to the collection. Renaming it as residue-ppi:A-B:1ATN_3
Query with ID residue-ppi:A-B:1ATN has already been added to the collection. Renaming it as residue-ppi:A-B:1ATN_4
```
because of the same base name ```1ATN```.

Now, the duplicate pdb file name will be checked using the standard below:
1. If the pdb file **contains an underscore character**, then it will further check whether the complete pdb file name(base name + name after underscore) already exists in the ```self._queries``` list.
For example for ```1ATN_1w.pdb``` contains an underscore, it will be checked if there exists a complete same query id in the query list named ```1ATN_1w``` instead of only checking whether the base name ```1ATN```exists.
2. If the pdb file **does not contain an underscore character**, it will check whether the complete query name(base name) already exists in ```self.ids_count``` list.
For example for ```1A6B.pdb```, which contains no underscore, if a same pdb file named ```1A6B.pdb``` appears, it will be renamed to ```1A6B_2```


joyceljy:
@DaniBodor  I am not sure if I'm understanding the issue correctly, can you give it a look?
If everything is fine then I will add a unit test for this later on.

deeprankcore/query.py
```diff
@@ -189,15 +189,27 @@ def add(self, query: Query, verbose: bool = False, warn_duplicate: bool = True):
             _log.info(f'Adding query with ID {query_id}.')
 
         query_id_base = query_id.split("_")[0]
+        
+        warn_duplicate = False
```
gcroci2, line 193: `warn_duplicate` is a parameter for the user to determine if printing warnings in case of duplicates or not. As you modified it, warnings will be printed independently from what the user has set up.

gcroci2 (changes requested):
Thanks for picking this up :)
I think we can solve the issue in a much simpler way, like the following:
```python
    def add(self, query: Query, verbose: bool = False, warn_duplicate: bool = True):
        """
        Adds a new query to the collection.

        Args:
            query(:class:`Query`): Must be a :class:`Query` object, either :class:`ProteinProteinInterfaceResidueQuery` or
                :class:`SingleResidueVariantAtomicQuery`.    
            verbose(bool, optional): For logging query IDs added, defaults to False.
            warn_duplicate (bool): Log a warning before renaming if a duplicate query is identified.

        """
        query_id = query.get_query_id()

        if verbose:
            _log.info(f'Adding query with ID {query_id}.')

        if query_id not in self.ids_count:
            self.ids_count[query_id] = 1
        else:
            self.ids_count[query_id] += 1
            new_id = query.model_id + "_" + str(self.ids_count[query_id])
            query.model_id = new_id
            
            if warn_duplicate:
                _log.warning(f'Query with ID {query_id} has already been added to the collection. Renaming it as {query.get_query_id()}')

        self._queries.append(query)
```
Here I am just directly using `query_id` to keep track of the queries ids in the `ids_count` dictionary, without splitting anything on `_` character. 
I am not sure why we didn't implement this in the first place, and why it was necessary to split on `_`, but I think it should work. Tests pass on my local machine, but this is not a guarantee since we don't have relevant tests yet, but at least it means that it doesn't break anything. 

Then as you mentioned, we need at least a unit test for testing relevant cases.

joyceljy:
Thanks, Giulia!
I am also aware that the ```ids_count``` on counting on the base_name of the pdb files and I thought that it may be used somewhere else. That's the reason why I chose not to modify that part. But like you said if there is no limitation on the name storing to ids_count, your way will be the most efficient way:)

tests/test_querycollection.py
```diff
@@ -219,3 +219,49 @@ def test_querycollection_process_combine_output_false():
...
+            chain_id1 = chain_id1,
+            chain_id2 = chain_id2,
+            targets = targets,
+            pssm_paths = {
+                chain_id1: pssm_path1,
+                chain_id2: pssm_path2
+            }
+        ))
+    duplicate_pdb = "1ATN_1w"
+    duplicate_pdb_id = "residue-ppi:" + chain_id1 + '-' + chain_id2 + ':' + duplicate_pdb
```
gcroci2, line 254: No need of such a complex thing to follow, you can retain the id by using the query attributes (should be `query.model_id`)

tests/test_querycollection.py
```diff
@@ -219,3 +219,49 @@ def test_querycollection_process_combine_output_false():
         assert len(output_paths) == cpu_count
 
         rmtree(output_directory)
+
+def test_querycollection_add():
```
gcroci2, line 223:
```suggestion
def test_querycollection_duplicates_add():
```

tests/test_querycollection.py
```diff
@@ -219,3 +219,49 @@ def test_querycollection_process_combine_output_false():
         assert len(output_paths) == cpu_count
 
         rmtree(output_directory)
+
```
gcroci2, line 222:
```suggestion


```

tests/test_querycollection.py
```diff
@@ -219,3 +219,49 @@ def test_querycollection_process_combine_output_false():
...
+    Tests add method of QueryCollection class.
+    """
+    ref_path = "tests/data/ref/1ATN/1ATN.pdb"
+    pssm_path1 = "tests/data/pssm/1ATN/1ATN.A.pdb.pssm"
+    pssm_path2 = "tests/data/pssm/1ATN/1ATN.B.pdb.pssm"
+    chain_id1 = "A"
+    chain_id2 = "B"
+    pdb_paths = [
+        "tests/data/pdb/1ATN/1ATN_1w.pdb",
+        "tests/data/pdb/1ATN/1ATN_1w.pdb",
```
gcroci2, line 234:
```suggestion
        "tests/data/pdb/1ATN/1ATN_1w.pdb",
        "tests/data/pdb/1ATN/1ATN_1w.pdb",
```

gcroci2: Make sure it works with more than 2 duplicates

tests/test_querycollection.py
```diff
@@ -219,3 +219,49 @@ def test_querycollection_process_combine_output_false():
...
+        "tests/data/pdb/1ATN/1ATN_2w.pdb",
```
gcroci2, line 235:
```suggestion
        "tests/data/pdb/1ATN/1ATN_2w.pdb",
        tests/data/pdb/1ATN/1ATN_2w.pdb",
```

gcroci2: 2 duplicates case

tests/test_querycollection.py
```diff
@@ -219,3 +219,49 @@ def test_querycollection_process_combine_output_false():
...
+            pdb_path = pdb_path,
+            chain_id1 = chain_id1,
+            chain_id2 = chain_id2,
+            targets = targets,
+            pssm_paths = {
+                chain_id1: pssm_path1,
+                chain_id2: pssm_path2
+            }
+        ))
+    duplicate_pdb = "1ATN_1w"
```
gcroci2, line 253: From line 253, the test is unnecessarily complex. 

You could do something like this:
```python
model_ids = []
for query in queries.queries:
    model_ids.append(query.model_id)

model_ids.sort()
```

And then few simpler assers:
```python
assert model_ids == ['1ATN_1w', '1ATN_1w_2', '1ATN_1w_3', '1ATN_2w', '1ATN_2w_2', '1ATN_3w']
assert queries.ids_count['residue-ppi:A-B:1ATN_1w'] == 3
assert queries.ids_count['residue-ppi:A-B:1ATN_2w'] == 2
assert queries.ids_count['residue-ppi:A-B:1ATN_3w'] == 1
```

DaniBodor (approved):
Looks really good! Elegant solution.

Could you just add an empty line to the end of the test file. No super important reason, but it's [recommended](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file) to always end on an empty line.

DaniBodor (approved):

DaniBodor:
I think merging is still blocked because @gcroci2 once requested changes and hasn't formally accepted yet. Can you please do so now :)

gcroci2 (approved):

Issue #451, opened 2023-06-23
Improve installation instructions


DaniBodor:
I recently had to reinstall the package on a fresh WSL install and ran into a few snafus that should be fixed or mentioned in the instructions:

- [x] The header of installing pytorch (below dssp) disappeared (I believe this is solved in #434)
- [x] `sudo apt-get install dssp` could give an error "E: Unable to locate package dssp"
  - solve by first running `sudo apt-get update`
- [x] `python -c "import torch; print(torch.version.cuda)"` could return "None", in which case "cpu" should be used.
- [x] if gcc isn't installed, there will be an error when pip installing the package itself. This can be addressed as follows:
  1. test whether gcc is installed: if `gcc --version` does not give an error: you're good
  2. if there is an error, run: `sudo apt-get install gcc` before pip installing
- [x] it would be nice if there were a test command or something that people could run to ensure that installation works as intended.
  - perhaps by running the integration tests?
- [x] In the editable install instructions, maybe it would be nice to list the extras packages for testing, etc?
  - Can we collapse all of the extras into a single option (dev?) and/or get rid of the ones we don't need anymore?
 
Maybe this can be resolved together with #450

Issue #450, opened 2023-06-23
Fix colliding paths


DaniBodor:
I just cloned the repo (on windows) and got the following warning:

> warning: the following paths have collided (e.g. case-sensitive paths
> on a case-insensitive filesystem) and only one from the same
> colliding group is in the working tree:
> 
>   'tests/data/pdb/1CRN/1CRN.pdb'
>   'tests/data/pdb/1crn/1CRN.pdb'

I'm pretty sure this is just duplicate data and one of the two can be removed. This should be confirmed, and if true one can be deleted (make sure that all tests point to the remaining data).

Issue #432, opened 2023-05-25
Add tutorials for PPIs


gcroci2:
For PPI, we need at least the following tutorials:
-  data generation
  - [x] residue
  - [x] atom
-  training
  - [x] classification for GNNs
  - [x] classification for CNNs
  - [x] references to regression

PR #452, opened 2023-06-27
docs: improve installation instructions

DaniBodor:
Added some more details to installation instructions. This is mainly stuff that could go wrong for super fresh systems.

I made the PR point to #432, since you also made changes to README there, plus I am referring to the tutorials.

closes: #451 and #450


DaniBodor:
closes #450 and #451

README.md
```diff
@@ -66,31 +67,44 @@ Before installing deeprankcore you need to install some dependencies. We advise
   - Here, `${TORCH}` and `${CUDA}` should be replaced by the pytorch and CUDA versions installed. You can find these using:
     - `python -c "import torch; print(torch.__version__)"` and
     - `python -c "import torch; print(torch.version.cuda)"`
+      - if this returns `None`, use `cpu` instead
   - For example: `https://data.pyg.org/whl/torch-2.0.0+cpu.html`
-* Only if you have a MacOS with M1 chip, run also:
-  * `conda install pytables`
+* Check if gcc is installed: `gcc --version`.
```
gcroci2, line 72: For which dependency is this needed? I had it already on my mac so I am not sure

README.md
```diff
@@ -66,31 +67,44 @@ Before installing deeprankcore you need to install some dependencies. We advise
...
+```bash
 git clone https://github.com/DeepRank/deeprank-core
 cd deeprank-core
 pip install -e ./
 ```
 
-## Documentation
+## Test installation
+
+To check that all components were installed correctly, run either of the following commands and ensure that all tests pass.
```
gcroci2, line 94: This is possible if the user has cloned the repo only, otherwise, there will be no tests/ folder. I'd just move this paragraph below the installation using the cloned repo. Also theoretically if the user installs the package from the version deployed on pypi he/she shouldn't care about testing that the package works (because it should); I know I specified the testing part in the tutorials but indeed I was a bit uncertain about that

docs/installation.md
```diff
@@ -2,38 +2,50 @@
...
+  * CPU only: `conda install pytorch==2.0.0 torchvision==0.15.0 torchaudio==2.0.0 cpuonly -c pytorch`
+  * if using GPU: `conda install pytorch==2.0.0 torchvision==0.15.0 torchaudio==2.0.0 pytorch-cuda=11.7 -c pytorch -c nvidia`
+* [pytorch-geometric](https://pytorch-geometric.readthedocs.io/en/latest/notes/installation.html): `conda install pyg -c pyg`
+* [Dependencies for pytorch geometric from wheels](https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html#installation-from-wheels): `pip install pyg_lib torch_scatter torch_sparse torch_cluster torch_spline_conv -f https://data.pyg.org/whl/torch-${TORCH}+${CUDA}.html`. 
+  - Here, `${TORCH}` and `${CUDA}` should be replaced by the pytorch and CUDA versions installed. You can find these using:
+    - `python -c "import torch; print(torch.__version__)"` and
+    - `python -c "import torch; print(torch.version.cuda)"`
+      - if this returns `None`, use `cpu` instead
+  - For example: `https://data.pyg.org/whl/torch-2.0.0+cpu.html`
+* Check if gcc is installed: `gcc --version`.
```
gcroci2, line 22: Same comment as for the readme, in case you need to add info just to the same here

docs/installation.md
```diff
@@ -2,38 +2,50 @@
...
-```
+```bash
 git clone https://github.com/DeepRank/deeprank-core
 cd deeprank-core
 pip install -e ./
 ```
+
+## Test installation
+
+To check that all components were installed correctly, run either of the following commands and ensure that all tests pass.
```
gcroci2, line 44: Same comment as for the readme, in case you need to add info just to the same here

gcroci2 (approved):
Nice additions and improvements :) I added very minor comments

README.md
DaniBodor: I think for freesasa, if I remember correctly

gcroci2: indeed I also think it's not necessary

PR #454, opened 2023-06-27
CI: fewer actions

DaniBodor:
Changes to workflows:
- don't run actions on draft PRs
- only run actions if relevant files were modified
- cache environment to run actions on (not working yet).

closes #453

PR #457, opened 2023-07-04
CI: fewer triggers

### LINKED ISSUE #453: Reduce environmental impact of CI
See this [blog post](https://blog.esciencecenter.nl/reduce-reuse-recycle-save-the-planet-one-github-action-at-a-time-4ab602255c3f).

DaniBodor:
Changes to workflows to reduce environmental impact of CI (see [here](https://blog.esciencecenter.nl/reduce-reuse-recycle-save-the-planet-one-github-action-at-a-time-4ab602255c3f)):
- don't run actions on draft PRs
  - but activated when hitting "ready for review"
- only run actions if relevant files/folders were modified
- only run CI on python 3.9
  - if we really also want 3.10 CI, then maybe it would make sense to do coveralls in 3.9 and build in 3.10, seeing that coveralls actually runs building as well.

closes #453 
redo of #454


.github/workflows/build.yml
```diff
@@ -1,25 +1,49 @@
 name: build
 
 on:
-  # Trigger analysis when pushing in main or pull requests, and when creating
-  # a pull request. 
   push:
+    paths:
+      # specific folder locations
+      - "deeprankcore/**"
```
gcroci2, line 7: Instead of listing all paths, does it work if we list only the ones we want to exclude? If yes, I'd go for that. Less verbose and easier to maintain. Same for the others yml

.github/workflows/build.yml
```diff
@@ -1,25 +1,49 @@
...
+    if: github.event.pull_request.draft == false
     name: Build for (${{ matrix.python-version }}, ${{ matrix.os }})
     runs-on: ${{ matrix.os }}
 
     strategy:
       fail-fast: false
       matrix:
         os: ['ubuntu-latest']
-        python-version: ['3.9', '3.10']
+        python-version: ['3.9']
```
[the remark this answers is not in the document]
DaniBodor: Newest python is actually 3.11 :). But let's stick to 3.9 and 3.10 for now, I think.
We could do 3.9 for the normal build and 3.10 for the coveralls build. There is currently 1 command present in the normal build that isn't in coveralls, I can add that one to the coveralls to basically make it an extension of normal build.

.github/workflows/build.yml
DaniBodor: why is that less verbose? I think there's many more file types and locations to exclude than there are to include.
I can try to find out if it's possible to maintain a single list of locations to include so we can refer to it in each yml file rather than listing them each time (although I don't really want to spend a lot of extra time on this).

DaniBodor: I guess ignore is safer, cause it's less likely we list something there that shouldn't be ignored as opposed to forgetting to add something to the current list that should not be ignored.

.github/workflows/build.yml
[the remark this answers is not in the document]
gcroci2: Oh and it was released in October, so many months ago. Then in the early future we should think about adding 3.11 as well, and at that point we can evaluate how critical is to remove the 3.9 version. 

For the building, I get that coveralls does the build itself so you want to test 3.9 with the build file and 3.10 with the coveralls file, because both builds are tested this way. But this is not the cleaner way of doing it, because these yml files are thought to have well separated functions and they originate the badges that we show in the readme:
image

The build should be verified separately from coveralls, which should be run only if the build jobs go well. There are ways or creating interdependent yml files and avoiding running stuff multiple times, but I didn't want to spend much time on implementing more elegant solutions and I sticked to the basic one. If you're interested of course go for it, and we can have a chat for looking together where to find such info, but otherwise I'd prefer to use the build file to actually verify the builds that we want to guarantee (3.09 and 3.10 so far), and then use one of the two versions for running coveralls.

gcroci2 (commented):
Very nice improvements :) and I agree that path-ignore is safer here. I only have a comment about removing the two builds from the build file, let me know what you want to do about it. For the rest, consider the PR approved!

gcroci2 (approved):

Issue #463, opened 2023-08-04
HDF5 file not generated when `exposure` feature module is used on PDB files from Propedia database


DanLep97:
**Describe the bug**
When building the HDF5 file of the graph database using exposure component from the propedia database (and protCID database as well), I get the following error:
```
"""
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/usr/local/lib/python3.9/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/root/deeprankcore/deeprank-core/deeprankcore/query.py", line 245, in _process_one_query
    graph.write_to_hdf5(output_path)
  File "/root/deeprankcore/deeprank-core/deeprankcore/utils/graph.py", line 218, in write_to_hdf5
    node_features_group.create_dataset(
  File "/usr/local/lib/python3.9/site-packages/h5py/_hl/group.py", line 183, in create_dataset
    dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
  File "/usr/local/lib/python3.9/site-packages/h5py/_hl/dataset.py", line 86, in make_new_dset
    tid = h5t.py_create(dtype, logical=1)
  File "h5py/h5t.pyx", line 1664, in h5py.h5t.py_create
  File "h5py/h5t.pyx", line 1688, in h5py.h5t.py_create
  File "h5py/h5t.pyx", line 1748, in h5py.h5t.py_create
TypeError: Object dtype dtype('O') has no native HDF5 equivalent
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/mnt/netcache/data/dlepikhov/propedia_ssl/script/build_propedia.py", line 44, in <module>
    h5_p = queries.process(
  File "/root/deeprankcore/deeprank-core/deeprankcore/query.py", line 329, in process
    pool.map(pool_function, self.queries)
  File "/usr/local/lib/python3.9/multiprocessing/pool.py", line 364, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/usr/local/lib/python3.9/multiprocessing/pool.py", line 771, in get
    raise self._value
TypeError: Object dtype dtype('O') has no native HDF5 equivalent
```
This happens only when trying to calculate the exposure feature component. At first, I thought it is because H atoms are a problem, but removing them from pdb files didn't help.

**Environment:**
[...]

PR #465, opened 2023-08-07
fix: cast `hse` feature to float64

DanLep97:
PR for #463


gcroci2 (approved):
Good that you noticed the error! Thanks for having fixed it :)

Issue #477, opened 2023-09-04
Investigate scipy 1.11.2 error


gcroci2:
If we use the latest release of scipy, 1.11.2 (in the CI it's now fixed at 1.10.0), tests fail:

```bash
============================================================================================================ short test summary info =============================================================================================================
FAILED tests/test_trainer.py::TestTrainer::test_dataset_equivalence_pretrained - numpy.linalg.LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
FAILED tests/test_trainer.py::TestTrainer::test_fout - numpy.linalg.LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
FAILED tests/test_trainer.py::TestTrainer::test_ginet - numpy.linalg.LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
FAILED tests/test_trainer.py::TestTrainer::test_ginet_class - numpy.linalg.LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
FAILED tests/test_trainer.py::TestTrainer::test_ginet_sigmoid - numpy.linalg.LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
FAILED tests/test_trainer.py::TestTrainer::test_incompatible_pretrained_no_Net - numpy.linalg.LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
FAILED tests/test_trainer.py::TestTrainer::test_incompatible_pretrained_no_test - numpy.linalg.LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
FAILED tests/test_trainer.py::TestTrainer::test_naive - numpy.linalg.LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
FAILED tests/test_trainer.py::TestTrainer::test_no_valid_full_train - numpy.linalg.LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
FAILED tests/test_trainer.py::TestTrainer::test_no_valid_provided - numpy.linalg.LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
FAILED tests/test_trainer.py::TestTrainer::test_sgat - numpy.linalg.LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
```

[...]

PR #482, opened 2023-09-04
fix: make scipy 1.11.2 work

cbaakman:
This error is described in #477. Converting the sparse array to a dense array type stops the error from appearing.


gcroci2 (approved):

gcroci2:
Let's wait for all the checks to pass, then we can merge

PR #487, opened 2023-09-06
refactor: query classes and related code

DaniBodor:


DaniBodor:
commits were rebased and will be opened as new PR

Issue #456, opened 2023-07-04
Remove h5explorer from package


DaniBodor:
This form of visualization isn't used or documented anywhere, and as of #434 has already been removed from setup.cfg.

Can be removed from github action and from all modules using it.

DaniBodor:
@gcroci2 , here is the issue

PR #491, opened 2023-09-06
build: improve installation making use of `pyproject.toml` file only and `setuptools`

gcroci2:
- I removed redundancies from the actions 
- Now we have only one config file instead of 3 (pyproject.toml), which contains all the needed info for pip-related dependencies and metadata of deeprank2.
- I fixed the builds for multiple python versions; the action.yml file was not taking as input the python version specified in the workflows, thus defaulting always to the one specified (as a default) in the action.yml, which was python 3.9. We were not running workflows on python 3.10.
- Now we test the build on Python 3.10 and 3.11; I removed python 3.9 from everywhere, so we're not maintaining it anymore.
- I kept the conda installation of `msms` since `ResidueDepth` from biopython uses it (tried and verified).
- [Apparently](https://stackoverflow.com/questions/68809295/extra-python-package-index-url-with-setuptools-now-that-dependency-links-is-depr) dependencies' urls (or index urls) are not expected from setuptools: the user should be in charge of choosing the indexes, and not the creator of the package; installing torch, torch_geometric and the other torch adds is not possible using setuptools beacause their installation needs such links (or conda, which is not supported by the toml project file). I've read also [here](https://github.com/pypa/setuptools/discussions/3860) that setuptools does not support such links, and that they don't comply to the new pep specifications. With poetry this was possible, but being not in line with the standards and having chosen to not use poetry for installing the other dependencies of this package, I think we should just instruct the user to install the pytorchs versions suitable for the machine, and we should not be responsible for that. As developers, we'll use actions/install-python-and-package/action.yml as a reference for how to install torch packages.
- Same as above for the other external dependencies. Also, is a pain to maintain the instructions update in the readme, so I just left the links to the right installation pages for all the external packages (i.e., packages not installed through the toml file). This makes our readme instructions much more readable, and avoids us to keep updating the readme each time one external dependency changes something (see now, MSMS removed the download file from the website). Eventually, we can update the links for pointing to more up-to-date installation pages, when and if they'll change. Thus, I updated the installation docs. 
- I removed h5xplorer from everywhere. This PR closes issue #456 as well. 
- I tried to setup MacOS installation, and it was running correctly (at least for Python 3.10) except for MSMS installation, which for Mac can't be via conda. [I opened an issue](https://github.com/SBRG/ssbio/issues/55) about it  on their repo, in case they solve it I will readd the MacOS installation. For now I left the OS checking in the action.yml, and in case it will be ready to be run.


pyproject.toml
DaniBodor, line 1: also haven't looked at this file in detail...

README.md
```diff
@@ -46,35 +46,22 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
-* Check if gcc is installed: `gcc --version`.
-  * if this gives an error, run `sudo apt-get install gcc`.
-
-* For MacOS with M1 chip (otherwise ignore this): `conda install pytables`
+Before installing deeprank2 you need to install some dependencies. We advise to use a [conda environment](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html) with Python >= 3.9 installed. Follow the official documentation linked below:
+
+*  [MSMS](https://ssbio.readthedocs.io/en/latest/instructions/msms.html)
+  *  [Here](https://ssbio.readthedocs.io/en/latest/instructions/msms.html) for MacOS with M1 chip users
+*  [PyTorch](https://pytorch.org/get-started/locally/)
+  *  We support torch's CPU library as well as CUDA
```
DaniBodor, line 60: This didnt render as an indent before (I think it will now)
```suggestion
    *  We support torch's CPU library as well as CUDA
```

README.md
```diff
@@ -46,35 +46,22 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
-      * If the package cannot be located, first run `sudo apt-get update`.
-    * on older versions of ubuntu or on mac or lacking sudo priviliges: install from [here](https://github.com/pdb-redo/dssp), following the instructions listed.
-* Check if gcc is installed: `gcc --version`.
-  * if this gives an error, run `sudo apt-get install gcc`.
-
-* For MacOS with M1 chip (otherwise ignore this): `conda install pytables`
+Before installing deeprank2 you need to install some dependencies. We advise to use a [conda environment](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html) with Python >= 3.9 installed. Follow the official documentation linked below:
+
+*  [MSMS](https://ssbio.readthedocs.io/en/latest/instructions/msms.html)
+  *  [Here](https://ssbio.readthedocs.io/en/latest/instructions/msms.html) for MacOS with M1 chip users
```
DaniBodor, line 58: didn't render as an indent. double check that it does now.

```suggestion
    *  [Here](https://ssbio.readthedocs.io/en/latest/instructions/msms.html) for MacOS with M1 chip users
```

README.md
```diff
@@ -46,35 +46,22 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
-    * on ubuntu 22.04 or newer: `sudo apt-get install dssp`.
-      * If the package cannot be located, first run `sudo apt-get update`.
-    * on older versions of ubuntu or on mac or lacking sudo priviliges: install from [here](https://github.com/pdb-redo/dssp), following the instructions listed.
-* Check if gcc is installed: `gcc --version`.
-  * if this gives an error, run `sudo apt-get install gcc`.
-
-* For MacOS with M1 chip (otherwise ignore this): `conda install pytables`
+Before installing deeprank2 you need to install some dependencies. We advise to use a [conda environment](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html) with Python >= 3.9 installed. Follow the official documentation linked below:
+
+*  [MSMS](https://ssbio.readthedocs.io/en/latest/instructions/msms.html)
```
DaniBodor, line 57: link to anaconda instead (or both??)

```suggestion
*  [MSMS](https://anaconda.org/bioconda/msms)
```

README.md
```diff
@@ -84,12 +71,12 @@ Once the dependencies are installed, you can install the latest stable release o
...
 
-Alternatively, get all the new developments by cloning the repo and installing the code with
+Alternatively, get all the new developments by cloning the repo and installing the editable version of the package with
 
 ```bash
 git clone https://github.com/DeepRank/deeprank2
 cd deeprank2
-pip install -e ./
+pip install -e .
 ```
```
DaniBodor, line 80: it would be nice to also include the command for adding the test packages as well (I always forget the where it needs to go and which part needs to be in quotations): `pip install -e '.[test]'`

README.md
```diff
@@ -46,35 +46,22 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
-  * if this gives an error, run `sudo apt-get install gcc`.
-
-* For MacOS with M1 chip (otherwise ignore this): `conda install pytables`
+Before installing deeprank2 you need to install some dependencies. We advise to use a [conda environment](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html) with Python >= 3.9 installed. Follow the official documentation linked below:
+
+*  [MSMS](https://ssbio.readthedocs.io/en/latest/instructions/msms.html)
+  *  [Here](https://ssbio.readthedocs.io/en/latest/instructions/msms.html) for MacOS with M1 chip users
+*  [PyTorch](https://pytorch.org/get-started/locally/)
+  *  We support torch's CPU library as well as CUDA
+*  [PyG](https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html) and its optional dependencies: `torch_scatter`, `torch_sparse`, `torch_cluster`, `torch_spline_conv`
```
DaniBodor, line 61: consider adding them to dependencies instead?

DaniBodor (approved):
Great job, big effort!

PR should be renamed as you are not using poetry in the end. Also, should it be a "build" type rather than a "refactor"?

I could follow the install instructions and could run pytest and prospector without anything getting flagged.
I did notice that somewhere along the way scipy 1.11.1 gets installed, then gets uninstalled and 1.11.2 gets installed. Probably not worthwhile investigating further though.

I left some minor comments here and there.

gcroci2:
> Great job, big effort!
> 
> PR should be renamed as you are not using poetry in the end. Also, should it be a "build" type rather than a "refactor"?
> 
> I could follow the install instructions and could run pytest and prospector without anything getting flagged. I did notice that somewhere along the way scipy 1.11.1 gets installed, then gets uninstalled and 1.11.2 gets installed. Probably not worthwhile investigating further though.
> 
> I left some minor comments here and there.

I changed the PR name :) regarding scipy, I think it gets installed with torch and then when we install the package via the toml file it gets upgraded because of the requirement there (>= 1.11.2), which is needed. I think it's fine for now

README.md
gcroci2: Done, but JFYTK you can look at the action.yml file, and there you'll see how the test dependencies are added :)

README.md
gcroci2: I remembered why I didn't do it at the end. As you can see [here](https://pytorch.org/), installing torch via pip requires an --index-url; same for pytorch geometric additional deps: https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html. Could be that we're lucky and pip picks the right things for ubuntu (even it would be weird but still could be), but I'd stick to the official docs which require urls, that can't be provided in the toml file if we're using setuptools. Also in the future it will be easier to add macos build since there pytorch and pytorch geometric can't be installed via pip.

README.md
```diff
@@ -50,18 +50,24 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
 
-*  [MSMS](https://ssbio.readthedocs.io/en/latest/instructions/msms.html)
-  *  [Here](https://ssbio.readthedocs.io/en/latest/instructions/msms.html) for MacOS with M1 chip users
+*  [MSMS](https://anaconda.org/bioconda/msms): `conda install -c bioconda msms`.
+    *  [Here](https://ssbio.readthedocs.io/en/latest/instructions/msms.html) for MacOS with M1 chip users.
 *  [PyTorch](https://pytorch.org/get-started/locally/)
-  *  We support torch's CPU library as well as CUDA
-*  [PyG](https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html) and its optional dependencies: `torch_scatter`, `torch_sparse`, `torch_cluster`, `torch_spline_conv`
+    *  We support torch's CPU library as well as CUDA.
+*  [PyG](https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html) and its optional dependencies: `torch_scatter`, `torch_sparse`, `torch_cluster`, `torch_spline_conv`.
```
DaniBodor, line 63: you want to give the command for these as well? Just so users don't have to scroll the external sites to find it?

DaniBodor (approved):

README.md
gcroci2: Since both pytorch and pytorch geom have the nice table which the user can set according to the OS, and it gets update quite often, I'd leave to the user to find the commands, unless at a certain point it becomes necessary because a specific version is needed (right now all is working with the latest torch versions).

PR #493, opened 2023-09-06
docs: add performances table for deeprank2

gcroci2:
I am using the same PDB raw data used in the tutorials (see https://zenodo.org/record/8187806).

- Experiments done on Apple M1 Pro, 1 CPU used 
- [x] PPIs
  - atomic resolution, distance cutoff 5.5, grid resolution 1 (n points 35x30x30), 33 features (all except for `conservation` features module)
  - mean on 100 PDB files, values are per data point (one ppi)
  - graph: 2.9879 s (std 0.2260), 0.5387 MB (std 0.0706)
  - graph + grid: 11.3493 s (std 1.3049), 16.0929 MB (std 0.4414)
- [x] Variants
  - atomic resolution, distance cutoff 5.5, radius 10, grid resolution 1 (n points 35x30x30), 26 features (all except for `conservation`)
  - mean on 96 PDB files, values are per data point (one srv)
  - graph: 2.1989 s (std 0.0790), 0.0543 MB (std 0.0024)
  - graph + grid: 2.8514 s (std 0.1035), 17.5153 MB (std 0.5932)
- There are two scripts I used for calculating the computational performances for the table, and I think it would be useful to keep them. Not sure where thought. In tools/ maybe, or in a perf/ folder in tutorials? Or just in a branch we can reference. Not sure about this, @DaniBodor waiting for your thoughts :)
- I think it would be fair to compare with previous packages, but after thinking about how to do this I realized that it would take much more time than I would like to devote to this task now. I opened a new issue about it (#500) and I placed it in the new SS kanban board. For the future framework paper (not JOSS), it would be nice to have a proper computational performances comparison.


DaniBodor:
I will take a closer look at this, but for now already a few comments:

I feel that this should be much lower in the README, maybe just before the Package Development header. It's nice information to have there, but most users will not care that much, so it shouldn't be something they need to scroll past before getting to the information they are looking for.

> There are two scripts I used for calculating the computational performances for the table, and I think it would be useful to keep them. Not sure where thought. In tools/ maybe, or in a perf/ folder in tutorials? Or just in a branch we can reference. Not sure about this, @DaniBodor waiting for your thoughts :)

Yes, I would create a subfolder performance/, probably in tools.

> I think it would be fair to compare with previous packages, but after thinking about how to do this I realized that it would take much more time than I would like to devote to this task now. I opened a new issue about it (

Definitely out of the scope of this PR/the JOSS paper

> DSSP expects PDB files only and we have another format for variants

I think it's actually the same format, just a different extension. If you just replace the extension with .pdb, I think it should work. Not sure whether it's worth the effort though.

gcroci2:
> I feel that this should be much lower in the README, maybe just before the Package Development header. It's nice information to have there, but most users will not care that much, so it shouldn't be something they need to scroll past before getting to the information they are looking for.

Agreed, I moved it down

> > DSSP expects PDB files only and we have another format for variants
> 
> I think it's actually the same format, just a different extension. If you just replace the extension with .pdb, I think it should work. Not sure whether it's worth the effort though.

Indeed it worked :) I updated the results 

Regarding the location of the perf scripts, I created perf/ in tests/ and I placed them there. The reason why I don't want to put them into the package is that they are not tested (and doesn't make sense to test them in my opinion), and they will increase lines of code of the package while decreasing the overall coverage. Also they're not something that we wish is installed together with the package. I'm still not sure of the best place for them, but I'd say not in deeprank2/ code base folder.

README.md
```diff
@@ -315,6 +316,15 @@ trainer.test()
 
 ```
 
+## Computational performances
+
+The data generation process is rather efficient within DeepRank2. As an example, we show the time required to process the tutorials' PDB files available at [this](https://zenodo.org/record/8187806) Zenodo address. Atomic resolution and `distance_cutoff` of 5.5 Å have been used for both PPIs and SRVs, and the SRVs `radius` was set to 10 Å. The experiments are done on Apple M1 Pro, using 1 CPU only. Measures are computed on 100 data points for PPIs, and 96 for SRVs.
```
DaniBodor, line 321:
```suggestion
We measured the efficiency of data generation in DeepRank2 using the tutorials' [PDB files](https://zenodo.org/record/8187806) (+/- 100 data points per data set).
Below results were obtained as an average of X runs, on Apple M1 Pro, using a single CPU.
Parameter settings as follows: resolution: atomic; `distance_cutoff`:5.5 Å; radius (for SRV only): 10 Å. 
```

README.md
```diff
@@ -315,6 +316,15 @@ trainer.test()
...
+| PPIs | `components`, `contact`, `exposure`, `irc`, `secondary_structure`, `surfacearea`, 33 features in total. | `conservation` feature module was not used because PSSM files were not available for the data.                                                                                                                                                 | graph only: **3.58** (std 0.27) <br />graph+grid: **12.92** (std 1.39) | graph only: **0.54** (std 0.07) <br />graph+grid: **16.09** (std 0.44) |
+| SRVs | `components`, `contact`, `exposure`, `irc`, `secondary_structure`, `surfacearea`, 26 features in total.                        | Same as above for `conservation`. | graph only: **1.82** (std 0.37)   <br />graph+grid: **2.58** (std 0.66) | graph only: **0.05** (std 0.01)   <br />graph+grid: **8.52** (std 8.50)|
```
DaniBodor, line 326: I would remove the first two columns from the table and instead make a general statement about the features used above (especially because it's now the same for both).

tests/perf/ppi_perf.py
DaniBodor, line 1: It feels to me that these two scripts are better suited to be notebooks. That simplifies rerunning this with different parameter sets or something in the future, if needed. Or pointing it to different data paths.

Right I cannot run it from its current location, probably because the data path doesn't match the expected one

tests/perf/srv_perf.py
```diff
@@ -0,0 +1,120 @@
...
+from deeprank2.utils.grid import GridSettings, MapMethod
+
+aa_dict = {"ALA": alanine, "CYS": cysteine, "ASP": aspartate,
+           "GLU": glutamate, "PHE": phenylalanine, "GLY": glycine,
+           "HIS": histidine, "ILE": isoleucine, "LYS": lysine,
+           "LEU": leucine, "MET": methionine, "ASN": asparagine,
+           "PRO": proline, "GLN": glutamine, "ARG": arginine,
+           "SER": serine, "THR": threonine, "VAL": valine,
+           "TRP": tryptophan, "TYR": tyrosine
+           }
```
DaniBodor, line 30: I feel like we have this or similar dictionary in other places in the code base as well. Maybe we should create this dictionary in deeprank2/domain/aminoacidlist.py instead, and load it from there.

That would also make the long import in line 11-17 (and similar in other places) obsolete.

I guess this is a separate issue though (I created #503 )

tests/perf/srv_perf.py
```diff
@@ -0,0 +1,120 @@
...
+import numpy
+import pandas as pd
+
+from deeprank2.domain.aminoacidlist import (alanine, arginine, asparagine,
+                                            aspartate, cysteine, glutamate,
+                                            glutamine, glycine, histidine,
+                                            isoleucine, leucine, lysine,
+                                            methionine, phenylalanine, proline,
+                                            serine, threonine, tryptophan,
+                                            tyrosine, valine)
```
DaniBodor, line 17: how about

```suggestion
from deeprank2.domain import aminoacidlist as aa
```

and then calling them as `aa.serine`, etc below

tests/perf/ppi_perf.py
gcroci2: > It feels to me that these two scripts are better suited to be notebooks. That simplifies rerunning this with different parameter sets or something in the future, if needed. Or pointing it to different data paths.

Does it make a difference? I mean, you can edit paths in the .py file as you'd do in the .ipynb. These scripts are not meant for users at all, I wanted to keep them for us only if in the future we need to change something for computing the timings. 

> Right I cannot run it from its current location, probably because the data path doesn't match the expected one

If you didn't download the data from Zenodo and you didn't place them into the perf/ folder, it can't indeed work.

DaniBodor: yeah, just leave it.

tests/perf/srv_perf.py
DaniBodor: don't spend time on this either...

DaniBodor (approved):

README.md
```diff
@@ -318,14 +318,13 @@ trainer.test()
...
-| SRVs | `components`, `contact`, `exposure`, `irc`, `secondary_structure`, `surfacearea`, 26 features in total.                        | Same as above for `conservation`. | graph only: **1.82** (std 0.37)   <br />graph+grid: **2.58** (std 0.66) | graph only: **0.05** (std 0.01)   <br />graph+grid: **8.52** (std 8.50)|
+We measured the efficiency of data generation in DeepRank2 using the tutorials' [PDB files](https://zenodo.org/record/8187806) (~100 data points per data set), averaging the results run on Apple M1 Pro, using a single CPU.
+Parameter settings were: atomic resolution, `distance_cutoff` of 5.5 Å, radius (for SRV only) of 10 Å. The [features modules](https://deeprank2.readthedocs.io/en/latest/features.html) used were `components`, `contact`, `exposure`, `irc`, `secondary_structure`, `surfacearea`, for a total of 33 features for PPIs and 26 for SRVs.
```
DaniBodor, line 322:
```suggestion
Parameter settings were: atomic resolution, `distance_cutoff` of 5.5 Å, radius (for SRV only) of 10 Å. The [features modules](https://deeprank2.readthedocs.io/en/latest/features.html) used were `components`, `contact`, `exposure`, `irc`, `secondary_structure`, `surfacearea`, for a total of 33 features for PPIs and 26 for SRVs (the latter do not use `irc` features).
```

PR #497, opened 2023-09-12
docs: type hinting and docstrings in molstruct

DaniBodor:
This was really bugging me while refactoring query, so did a separate extra PR for it.


deeprank2/molstruct/atom.py
```diff
@@ -50,21 +49,19 @@ def __init__( # pylint: disable=too-many-arguments
...
     def __eq__(self, other) -> bool:
-        return (
-            isinstance(self, type(other))
-            and self._residue == other._residue
-            and self._name == other._name
-        )
+        if isinstance (other, Atom):
+            return (self._residue == other._residue
+                    and self._name == other._name)
+        return NotImplemented
```
gcroci2, line 55: We're not checking type hints with a tool (like mypy), but I guess that having here `bool` would give errors with the `NotImplemented` case

deeprank2/molstruct/pair.py
```diff
@@ -16,32 +17,26 @@ def __init__(self, item1: Any, item2: Any):
...
-            and self.item2 == other.item2
-            or self.item1 == other.item2
-            and self.item2 == other.item1
-        )
+    def __eq__(self, other) -> bool:
+        """Compare the pairs as sets, so the order doesn't matter."""
+        if isinstance(other, Pair):
+            return (self.item1 == other.item1 and self.item2 == other.item2
+                    or self.item1 == other.item2 and self.item2 == other.item1)
+        return NotImplemented
```
gcroci2, line 33: Same as above

deeprank2/molstruct/residue.py
```diff
@@ -32,58 +35,51 @@ def __init__(
...
-            and self._chain == other._chain
-            and self._number == other._number
-            and self._insertion_code == other._insertion_code
-        )
+        if isinstance(other, Residue):
+            return (self._chain == other._chain
+                    and self._number == other._number
+                    and self._insertion_code == other._insertion_code
+            )
+        return NotImplemented
```
gcroci2, line 43: Same as above

deeprank2/molstruct/residue.py
```diff
@@ -1,22 +1,25 @@
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
 import numpy as np
+
 from deeprank2.molstruct.aminoacid import AminoAcid
 from deeprank2.molstruct.structure import Chain
 from deeprank2.utils.pssmdata import PssmRow
 
+if TYPE_CHECKING:
```
gcroci2, line 9: Why do we need this if?

deeprank2/molstruct/residue.py
```diff
@@ -32,58 +35,51 @@ def __init__(
...
+    def chain(self) -> Chain:
         return self._chain
 
     @property
     def amino_acid(self) -> AminoAcid:
         return self._amino_acid
 
     @property
-    def atoms(self):
+    def atoms(self) -> list["Atom"]:
```
gcroci2, line 68: Why is Atom within quotation marks?

deeprank2/molstruct/residue.py
```diff
@@ -32,58 +35,51 @@ def __init__(
...
             return f"{self._number}{self._insertion_code}"
-
         return str(self._number)
 
     @property
     def insertion_code(self) -> str:
         return self._insertion_code
 
-    def add_atom(self, atom):
+    def add_atom(self, atom: "Atom"):
```
gcroci2, line 82: Okay I guess the quotation marks are because of the TYPE_CHECKING thing, but I'll leave this comment here as well just in case

deeprank2/molstruct/structure.py
```diff
@@ -1,13 +1,16 @@
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
 from deeprank2.utils.pssmdata import PssmRow
 
+if TYPE_CHECKING:
```
gcroci2, line 5: Just in case, same question as above here

deeprank2/molstruct/structure.py
```diff
@@ -16,7 +19,9 @@ def __init__(self, id_: Optional[str] = None):
         self._chains = {}
 
     def __eq__(self, other) -> bool:
-        return isinstance(self, type(other)) and self._id == other._id
+        if isinstance(other, PDBStructure):
+            return self._id == other._id
+        return NotImplemented
```
gcroci2, line 24: Just in case, NotImplemented thing also here

deeprank2/molstruct/structure.py
```diff
@@ -27,26 +32,24 @@ def __repr__(self) -> str:
     def has_chain(self, chain_id: str) -> bool:
         return chain_id in self._chains
 
-    def get_chain(self, chain_id: str):
+    def get_chain(self, chain_id: str) -> "Chain":
```
gcroci2, line 35: Why Chain is not under the `if TYPE_CHECKING:`?

gcroci2 (commented):
I left minor comments, more for me to understrand than for suggesting changes. I removed the duplicated comments, but in case we decide to change something we need to remember that it may be in multiple files.

deeprank2/molstruct/atom.py
DaniBodor: I found this here, when researching something else (https://stackoverflow.com/questions/63503512/python-type-hinting-own-class-in-method). Not sure where I can find the formal correct way to format `__eq__`, but the commentors on this post all seem to agree that this is the correct way to do it.

deeprank2/molstruct/residue.py
DaniBodor: This avoids circular imports. We cannot import `Residue` from atom.py _AND_ `Atom` from residue.py. Like this it only uses the import for type hint checking:

https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/

deeprank2/molstruct/residue.py
DaniBodor: yes. In python 3.11 it wouldn't be necessary. We could do `from __future__ import annotations` to avoid the quotation marks ([same link as above](https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/)) , but I'm not sure whether that has any other effects, so kept it like this.

Maybe importing annotations is nicer though, what do you think?

deeprank2/molstruct/structure.py
DaniBodor: because it comes from the same module, so doesn't need to be imported (in fact, can't be). It still needs the quotation marks, because it is only defined afterwards.

DaniBodor:
> I left minor comments, more for me to understrand than for suggesting changes. I removed the duplicated comments, but in case we decide to change something we need to remember that it may be in multiple files.

Sorry, I was actually meaning to leave references to some of these changes when I made the PR, but then forgot :)

I added them now in reply to your comments

DaniBodor:
3 additional changes:

1. I used `from __future__ import annotations` to avoid the quotation marks around some of the type hints
2. I improved docstrings and their formatting according to this convention: https://realpython.com/documenting-python-code/#class-docstrings.
3. I moved `SingleResidueVariant` into residue.py as they are closely related classes. 
    - I even wonder if it would make sense to make it a child class of `Residue` instead of taking the `Residue` as an attribute (but that would require more coding than I am currently willing to do for this).

You should be able to go through the commits one by one to see each change in isolation, which might make reviewing easier.

gcroci2 (approved):

deeprank2/molstruct/atom.py
gcroci2: You could run [mypy](https://realpython.com/lessons/type-checking-mypy/) to verify that the type hints are right

deeprank2/molstruct/structure.py
gcroci2: oh sorry I didn't notice that it was defined there

gcroci2:
> 3 additional changes:
> 
> 1. I used `from __future__ import annotations` to avoid the quotation marks around some of the type hints
> 2. I improved docstrings and their formatting according to this convention: https://realpython.com/documenting-python-code/#class-docstrings.
> 3. I moved `SingleResidueVariant` into residue.py as they are closely related classes.
>    
>    * I even wonder if it would make sense to make it a child class of `Residue` instead of taking the `Residue` as an attribute (but that would require more coding than I am currently willing to do for this).
> 
> You should be able to go through the commits one by one to see each change in isolation, which might make reviewing easier.
> 
> Note that because I have "automatic import sorting upon save" turned on, some additional imports are switched around as well.

Regarding 3., JFYTK, we're using the [Google Style convention](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) for docstrings.

DaniBodor:
> > 3 additional changes:
> > 
> > 1. I used `from __future__ import annotations` to avoid the quotation marks around some of the type hints
> > 2. I improved docstrings and their formatting according to this convention: https://realpython.com/documenting-python-code/#class-docstrings.
> > 3. I moved `SingleResidueVariant` into residue.py as they are closely related classes.
> >    
> >    * I even wonder if it would make sense to make it a child class of `Residue` instead of taking the `Residue` as an attribute (but that would require more coding than I am currently willing to do for this).
> > 
> > You should be able to go through the commits one by one to see each change in isolation, which might make reviewing easier.
> > Note that because I have "automatic import sorting upon save" turned on, some additional imports are switched around as well.
> 
> Regarding 3., JFYTK, we're using the [Google Style convention](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) for docstrings. Nice that you moved `SingleResidueVariant` into residue.py, makes more sense. Regarding the possibility of making it a child of `Residue`, that also makes sense, not sure how much it's worth to spend time on doing it

Thanks for the ref to google convention. It does still follow that as well. I wasn't sure whether the class itself should get a docstring or only the init method. But in both style guides, they each get a docstring.

PR #498, opened 2023-09-19
docs: clarify ppi scoring metrics and add doc strings and tests

gcroci2:
We have an already implemented function, `compute_ppi_scores`, that takes as input a decoy (ppi model that needs evaluation) and a reference pdb structure, considered the native one. It computes lrmsd, irmsd, fnat, dockq, binary, and capri_classes scores. I see two applications of such function:
- Adding the score we're interested in (e.g., dockq) to the HDF5 files, as a target. Then use regression (or classification in other scores' cases) task for training a supervised model for predicting the selected score (e.g., dockq). The application could be using the pre-trained model to be able to predict how good an (unlabeled) generated structure is.
- Evaluate a pool of generated structures for which we have the native ones, using one of the scores listed above. 

Do we need other features regarding this? @LilySnow 
I was evaluating if it's worth it to change `compute_ppi_scores` to a class @DaniBodor, but I think a function is perfectly fine, and a class would be unnecessary for the scope. What do you think?


LilySnow:
nice work! No, we do not need new metrics for this. The two applications are indeed correct. If we have time, we should test this using data from our DeepRank paper (Just the small capri data is enough. )

deeprank2/tools/target.py
```diff
@@ -83,22 +83,40 @@ def add_target(graph_path: Union[str, List[str]], target_name: str, target_list:
...
-def compute_targets(pdb_path: str, reference_pdb_path: str) -> Dict[str, Union[float, int]]:
+def compute_ppi_scores(pdb_path: str, reference_pdb_path: str) -> Dict[str, Union[float, int]]:
 
     """
-    Compute targets and outputs them as a dictionary.
-    For classification:
-       - binary (scalar value is expected to be either 0 or 1)
-       - capri_classes (scalar integer values are expected)
+    Compute structure similarity scores and return them as a dictionary.
+    Such measures have been developed for evaluating the quality of the PPI models produced by
```
DaniBodor, line 90: the first sentence of the docstring should be on the same line as the triple-quote and then an empty line before the longer explanation.

Also, I reworded it slightly (can't do a code suggestion for some reason:

```
    """Compute structure similarity scores and return them as a dictionary.
    
    These measures have been developed for evaluating the quality of the PPI models produced by
    computational methods (docking models), and all of them compare the structural similarity
    between the decoys (computationally generated structures) and the experimentally solved native
    structures. To calculate these measures, the interface between the two interacting protein molecules
    is defined as any pair of heavy atoms from the two molecules within 5Å of each other.

    - lmrsd (ligand root mean square deviation) is a float value calculated for the backbone of
        the shorter chain (ligand) of the model after superposition of the longer chain (receptor).
        Lower scores represent better matching than higher scores.
    - imrsd (interface rmsd) is a float value calculated for the backbone atoms of the interface residues 
        (atomic contact cutoff of 10Å) after superposition of their equivalents in the predicted complex (model)
        Lower scores represent better matching than higher scores.
    - fnat (fraction of native contacts) is the fraction of native interfacial contacts preserved in
        the interface of the predicted complex. The score is a float in the range [0, 1], where higher values
        respresent higher quality.
    - dockq (docking model quality) is a continuous quality measure for docking models that instead of classifying into different
        quality groups. It combines fnat, lmrs, and irms and yields a float score in the range [0, 1], where
        higher values respresent higher quality.    
    - binary (bool): True if the irmsd is lower than 4.0, meaning that the decoy is considered high quality
        docking model, otherwise False.
    - capri classification (int). The possible values are: 1 (high quality, irmsd < 1.0), 2 (medium, irmsd < 2.0), 
        3 (acceptable, irms < 4.0), 4 (incorrect, irmsd >= 4.0)
    See https://onlinelibrary.wiley.com/doi/abs/10.1002/prot.10393 for more details about capri_classes, lrmsd, irmsd, and fnat.
    See https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0161879 for more details about dockq.

    Args:
        pdb_path (str): Path to the decoy.
        reference_pdb_path (str): Path to the reference (native) structure.

    Returns: a dictionary containing values for lrmsd, irmsd, fnat, dockq, binary, capri_class.
    """
```

gcroci2:
> nice work! No, we do not need new metrics for this. The two applications are indeed correct. If we have time, we should test this using data from our DeepRank paper (Just the small capri data is enough. )

The code is tested and we're sure it works, so we don't strictly need to verify the previous results :) I'd prefer to use this directly for the application we proposed for this budget, which is ranking antibody-antigen docking models.

tests/tools/test_target.py
```diff
@@ -33,10 +35,42 @@ def test_add_target(self):
             os.remove(graph_path)
 
 
-    def test_compute_targets(self):
-        compute_targets("tests/data/pdb/1ATN/1ATN_1w.pdb", "tests/data/ref/1ATN/1ATN.pdb")
+    def test_compute_ppi_scores(self):
+        scores = compute_ppi_scores(
```
DaniBodor, line 39: thanks form actually testing it :)

DaniBodor (approved):
Nice job!

The docstring is a bit long here and I'm not sure this is the first place users should be expected to go to find out how to evaluate their models. Maybe it would be good to add something about this to the documentation as well?

Also, do we need something equivalent for SRV as well?

DaniBodor:
> I was evaluating if it's worth it to change `compute_ppi_scores` to a class @DaniBodor, but I think a function is perfectly fine, and a class would be unnecessary for the scope. What do you think?

I agree.

gcroci2:
> Also, do we need something equivalent for SRV as well?

Nope, I double checked with @LilySnow as well, all these metrics are very specific for single interfaces in decoy models. It gets more complex if you have multiple interfaces (more than two chains), but usually the interfaces are evaluated one by one in such cases, only one interface is selected.

deeprank2/tools/target.py
```diff
@@ -128,7 +146,7 @@ def compute_targets(pdb_path: str, reference_pdb_path: str) -> Dict[str, Union[f
     )
     scores[targets.BINARY] = scores[targets.IRMSD] < 4.0
 
-    scores[targets.CAPRI] = 5
+    scores[targets.CAPRI] = 4
     for thr, val in zip([6.0, 4.0, 2.0, 1.0], [4, 3, 2, 1]):
         if scores[targets.IRMSD] < thr:
```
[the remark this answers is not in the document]
gcroci2: Good catch :D

gcroci2:
Thanks for the suggestions :) @DaniBodor I added them. 

> The docstring is a bit long here and I'm not sure this is the first place users should be expected to go to find out how to evaluate their models. Maybe it would be good to add something about this to the documentation as well?

I agree, a doc string should definitely not contain all this information. I think it would be nice to have a .md section (as we did for docs/features.md) specific for docking, with a snippet code as an example and to move a big part of the doc string there. Something like this?

DaniBodor:
> I agree, a doc string should definitely not contain all this information. I think it would be nice to have a .md section (as we did for docs/features.md) specific for docking, with a snippet code as an example and to move a big part of the doc string there. Something like this?

yes, indeed.

gcroci2:
See [here](https://deeprank2.readthedocs.io/en/54_docking_metrics_gcroci2/docking.html) for the new page @DaniBodor. Consider that part of the SS budget is meant for creating tutorials for the new applications, including docking. So for now having something indicative is fine, later we can do proper notebooks with more details (and improved code, I don't like add_targets function at all). But since we'll work on docking tasks as well, we will develop an easier way to add targets to hdf5 files.

DaniBodor:
> See [here](https://deeprank2.readthedocs.io/en/54_docking_metrics_gcroci2/docking.html) for the new page @DaniBodor. Consider that part of the SS budget is meant for creating tutorials for the new applications, including docking. So for now having something indicative is fine, later we can do proper notebooks with more details (and improved code, I don't like add_targets function at all). But since we'll work on docking tasks as well, we will develop an easier way to add targets to hdf5 files. (I opened issue #499 about this)

Indeed, I was going to mention that it might be good to include a notebook/tutorial, but I agree that for now this is fine. Maybe there could be a short comment in the README about this, though, as currently it is completely hidden to users that this exists.

gcroci2:
> here

I added a ref in the readme and in the docs index to it. I'll wait for all the tests to pass and then I'll merge.

PR #504, opened 2023-09-22
feature: separate max edge distance from interaction radius throughout

### LINKED ISSUE #460: Clarify `distance_cutoff` across different `Query` classes
While developing tutorials for variants (PR #435) I noticed that the logic behind the `distance_cutoff` parameter across different query types is not very linear and clear, at least from my perspective. 

In variant query classes, the residues/atoms of interest are selected with the radius (via `get_surrounding_residues`) around the variant. 

In PPIs query classes, `distance_cutoff` is used for `_load_ppi_atoms` (I guess selecting the interface, within the distance cutoff).

Then, in all queries logic, `build_residue_graph` (or `build_atomic_graph`) builds a graph of the selected elements, and takes the distance cutoff as input, which in `build_residue_graph` is defined as the *shortest* interatomic distance between two residues.
- Actually, looking into `build_residue_graph`, then distances < the distance cut off are selected, making the cutoff distance actually the max and not the min. 
- `build_residue_graph` (and `build_atomic_graph`) selects distances < the distance cut off, but at least for PPIs this should have been done already in `_load_ppi_atoms` (and for variants in `get_surrounding_residues`). Can we remove this parameter from `build_residue_graph`?
- `Graph` class has `cutoff_distance` param as well, but again I don't think it's needed (never used in the `Graph` class).

Your feedback here would be very useful :) @DaniBodor

DaniBodor:
I think this one is also best reviewed one commit at a time, even though it's only 3 commits.
The final commit is not really part of this issue, but I went through the entire code and fixed/simplified/added type hinting throughout the code base. I can move it to a separate PR if you think it's better, but I had done it in between this one and #506, which builds on this. That is why it is currently here.

Changes:
- Clarify `distance_cutoff` vs `radius` across different `Query` classes (see [here](https://github.com/DeepRank/deeprank2/issues/460#issuecomment-1713910900))
    - Rename the parameters to avoid confusion
        - `radius` becomes `interaction_radius` = the distance from the structure of interest (variant residue or interface) to look for residues to include
        - `cutoff` becomes `max_edge_distance` = the maximum distance between two nodes to include an edge between them.
    - Add `interaction_radius` parameter for PPI; previously only SRVs were using separate paramenters, while PPI was using `cutoff` to mean both things.
    - Remove `edge_distance_cutoff` as a `Graph` attribute, because it serves no purpose.
- Decided against allowing for setting these to 0 to mean infinity/include all
   - It's quite a bit of effort, especially with adding tests (which would also be slow). 
   - It's very rare that someone should want to do this, so we should not make it "easy" to do so for users who don't know what they're doing.
   - If someone really wants to include everything, they can just be set to an extremely high value and it will do the same.

closes #460


gcroci2:
I've asked @cbaakman's review as well, only to verify that this PR doesn't break some logic that was meant for the old cutoffs. No need to go through the code in-depth, a conceptual confirmation would be more than enough :)

deeprank2/query.py
```diff
@@ -48,8 +48,10 @@ class Query:
         chain_ids (list[str] | str): the chain identifier(s) of the variant residue or interacting interfaces.
             Note that this does not limit the structure to residues from this/these chain(s).
         pssm_paths (dict[str, str]): the name of the chain(s) (key) and path to the pssm file(s) (value).
-        distance_cutoff (float): the maximum distance between two nodes to generate an edge connecting them.
         targets (dict[str, float]) = Name(s) (key) and target value(s) (value) associated with this query.
+        interaction_radius (float | None): all residues within this radius from the variant residue or interacting interfaces
+            will be included in the graph, irrespective of the chain they are on.
+        max_edge_distance (float | None): the maximum distance between two nodes to generate an edge connecting them.
```
gcroci2, line 54: I'd suggest `max_nodes_distance` (maximum distance between nodes for having an edge) or `max_distance_edge` (same but making nodes implicit) to avoid confusion from the naming

deeprank2/utils/buildgraph.py
```diff
@@ -209,16 +209,16 @@ def get_residue_contact_pairs( # pylint: disable=too-many-locals
...
-    """Get the residues that contact each other at a protein-protein interface.
+    """Find all residue pairs that may influence each other.
 
     Args:
         pdb_path (str): The path of the pdb file, that the structure was built from.
         structure (:class:`PDBStructure`): From which to take the residues.
         chain_id1 (str): First protein chain identifier.
         chain_id2 (str): Second protein chain identifier.
-        distance_cutoff (float): Max distance between two interacting residues.
+        interaction_radius (float): Maximum distance between residues to consider them as interacting.
```
gcroci2, line 220: since edges represent interactions, then isn't `interaction_radius` conceptually the same thing as `max_edge_distance`? Maybe the radius doesn't have to be necessarily about interactions (thus could be `radius` only), while edges should always be about interactions. Otherwise, what's the conceptual difference between these two variables?

gcroci2 (commented):
Nice work, almost there :) I left few comments we need to reflect on before proceeding with the merging. Let me know what do you think about them

deeprank2/utils/buildgraph.py
DaniBodor: Edges don't necessarily equate to interactions. I think mostly this will be the case, and by default I would use the same value (and we do), but there could be situations where `interaction_radius` is higher than `max_edge_distance`. This could be the case 1) purely to decrease computational power/time, or 2) e.g. because the "secondary" interactions can be relevant. Even if two nodes don't have an edge between them, an intermediate node could be affected by both edges.
I can't off the top of my head think of a situation where `max_edge_distance` would be larger than `interaction_radius`, and am not sure that would even make sense to do. Maybe we can create a waning/error/cap in case the user sets cutoff > radius. What do you think?

Note that in `interaction_radius` I am not necessarily talking about protein-protein-interactions (trans interactions), but could also be cis-interactions (so residues of the same protein interacting with each other). So also in a variant, if residues are too far away, they would not interact with/affect the variant residue so don't need to be considered. 

I agree that the terminology is confusing, though. At the same time, I find `radius` too generic, it doesn't really describe what the parameter is either. Let's see if we can come up with another nomenclature. Maybe something along the lines of `sphere_of_influence`, although that sounds too fuzzy to me as well.

deeprank2/query.py
DaniBodor: I agree that distance is not the correct terminology. How** about `max_edge_length`?

DaniBodor: it's a bit tough because "length" and "distance" both have a different meaning than this in [graph theory](https://www.geeksforgeeks.org/graph-measurements-length-distance-diameter-eccentricity-radius-center/)

deeprank2/utils/buildgraph.py
DaniBodor: as above, "radius" also has a specific meaning in graph theory.

DaniBodor:
So we just need to agree on a nomenclature here, and then I can merge this PR.

My newest suggestions are `influence_radius` and either `max_edge_length` or `edge_length_cutoff` (see below for they represent). What do you think?

> `radius` becomes `interaction_radius` = the distance from the structure of interest (variant residue or interface) to look for residues to include
> `cutoff` becomes `max_edge_distance` = the maximum distance between two nodes to include an edge between them.

deeprank2/query.py
gcroci2: max_edge_length

deeprank2/utils/buildgraph.py
gcroci2: influence_radius

gcroci2:
> So we just need to agree on a nomenclature here, and then I can merge this PR.
> 
> My newest suggestions are `influence_radius` and either `max_edge_length` or `edge_length_cutoff` (see below for they represent). What do you think?
> 
> > `radius` becomes `interaction_radius` = the distance from the structure of interest (variant residue or interface) to look for residues to include
> > `cutoff` becomes `max_edge_distance` = the maximum distance between two nodes to include an edge between them.

Indeed, so as agreed I'd go for `max_edge_length` and `influence_radius`

deeprank2/query.py
```diff
@@ -322,8 +327,10 @@ class ProteinProteinInterfaceQuery(Query):
         chain_ids (list[str] | str): the chain identifiers of the interacting interfaces (generally a single capital letter each).
             Note that this does not limit the structure to residues from these chains.
         pssm_paths (dict[str, str]): the name of the chain(s) (key) and path to the pssm file(s) (value).
-        distance_cutoff (float): the maximum distance between two nodes to generate an edge connecting them.
         targets (dict[str, float]) = Name(s) (key) and target value(s) (value) associated with this query.
+        interaction_radius (float | None): all residues within this radius from the interacting interfaces
```
gcroci2, line 331:
```suggestion
        interaction_radius (float | None): all residues within this radius from the interacting interface
```

gcroci2 (approved):

PR #513, opened 2023-10-10
docs: add pip install commands for the dependencies listed in the README

DarioMarzella:
Add pip install commands for the dependencies listed in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,9 +59,9 @@ Before installing deeprank2 you need to install some dependencies. We advise to 
 
 *  [MSMS](https://anaconda.org/bioconda/msms): `conda install -c bioconda msms`.
     * [Here](https://ssbio.readthedocs.io/en/latest/instructions/msms.html) for MacOS with M1 chip users.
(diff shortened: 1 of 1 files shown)


DaniBodor:
@gcroci2 removed the explicit commands on purpose. It is always better to conda install it rather than pip. Also, it changes to regularly, that we don't want to keep updating the README.
That being said, I also do like having the commands there, with a disclaimer saying "this is correct as of date X, if it doesn't work, check the link". Giulia, what do you think?

gcroci2:
I am still against this, the torch installations are very verbose and change too often. Also now, we should have `pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu` instead of the shorter `pip install torch` if you look at the official page. Also for geometric the official installation for pytorch geometric is different from the one you suggested @DaniBodor.

DaniBodor:
OK. Let's leave it as is then and close this PR without merging?

PR #507, opened 2023-09-23
refactor: unify buildgraph.py and `build_graph` functions for atom and residue level graphs

DaniBodor:
This PR is easiest reviewed one commit at a time.

These are the changes, in order of commits:

- creates central amino acid dictionaries instead of having them created multiple times at different parts throughout the code.
- improve formatting (line spacing, etc) and documentation (fewer comments, more docstrings, etc.) of buildgraph.py
- simplified code that reads atom data from pdb2sql objects
- simplified code that finds residues within `influence_radius`
- unify functions for building atomic and residue graphs (last 2 commits)

fix: #506, #503

blocked by: #492, #504

PR #522, opened 2023-11-08
linting: change linter to ruff

gcroci2:
Due to [the broken change of python extension in vscode](https://github.com/microsoft/vscode-python/wiki/Migration-to-Python-Tools-Extensions), we have to reset linter and formatter used in this project.

[Ruff](https://docs.astral.sh/ruff/) is both a linter and formatter, providing an all-in-one solutions with surprising performance(see [here1](https://astral.sh/ruff) and [here2](https://astral.sh/blog/the-ruff-formatter)). Ruff replaces prospector, and it has also isort functionalities.

⚠️ Note that developers have to reset their dev env by installing ruff and updating vscode extensions.

Inspired by @CunliangGeng 

- Blocked by #515, #504, #507 
- Run ruff as checker and formatter on the entire code-base, and fix issues
- Go through the ignore checks in the toml file, and evaluate which ones to remove (and fix the code accordingly)
- As mentioned [here](https://www.sicara.fr/blog-technique/boost-code-quality-ruff-linter#:~:text=Pylint%20and%20Ruff%20are%20often,doesn't%20really%20handle%20typing.), we should add `mypy` as a complement to Ruff (for typing rules).


DaniBodor:
replaced by #554

Issue #494, opened 2023-09-06
ValueErrors raised in Variant tutorials


DaniBodor:
I noticed that in processing the querycollection (for both atomic and residue), a number of ValueErrors are raised. The code doesn't crash though.

We should investigate why these are raised and whether it is ok that processing does not terminate when these are found.

gcroci2:
#613 is a duplicated of this issue. We will fix this in #613

PR #528, opened 2023-11-27
docs: add `Dockerfile` and `.yml` for building conda env

gcroci2:
The `Dockerfile` proposed here downloads the executable file from dssp repo, and installs it in the container context. It also uses a yml file (together with a txt file) for installing conda and pip dependencies, respectively. 

@DaniBodor for reviewing this PR you can follow the new docker instructions and the new installation instructions.

I opened a separate issue for creating an action for testing the Dockerfile #529.

When this PR is going to be merged, remember to update the JOSS' reviewers (https://github.com/openjournals/joss-reviews/issues/5983)


README.md
```diff
@@ -37,42 +37,81 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
+    - [GraphDataset](#graphdataset)
+    - [GridDataset](#griddataset)
+  - [Training](#training)
   - [Computational performances](#computational-performances)
   - [Package development](#package-development)
 
 ## Installation
 
-The package officially supports ubuntu-latest OS only, whose functioning is widely tested through the continuous integration workflows. 
+Note that the package officially supports ubuntu-latest OS only, whose functioning is widely tested through the continuous integration workflows.
```
DaniBodor, line 55: I think it would be nice to give a short summary of the different options here.

```suggestion
Note that the package officially supports ubuntu-latest OS only, whose functioning is widely tested through the continuous integration workflows. 

You can either install DeepRank2 in a [dockerized container](#dockerfile), which will allow you to test the package and run our [tutorial notebooks](https://github.com/DeepRank/deeprank2/tree/main/tutorials), or you can [install the full version](...) of DeepRank2.
```

README.md
```diff
@@ -37,42 +37,81 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
 ## Installation
 
-The package officially supports ubuntu-latest OS only, whose functioning is widely tested through the continuous integration workflows. 
+Note that the package officially supports ubuntu-latest OS only, whose functioning is widely tested through the continuous integration workflows. 
 
-### Dependencies
+### Dockerfile
 
-Before installing deeprank2 you need to install some dependencies. We advise to use a [conda environment](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html) with Python >= 3.10 installed. The following dependency installation instructions are updated as of 14/09/2023, but in case of issues during installation always refer to the official documentation which is linked below:
+In order to try out the package without worrying about your OS and without the need of installing all the required dependencies, we created a `Dockerfile` that can be used for taking care of everything in a suitable container. After having cloned the repository and installed [Docker](https://docs.docker.com/engine/install/), run the following commands from the root of the repository.
```
DaniBodor, line 59: what are the downsides of this, if any? Why wouldn't people always do this rather than go through the individual steps described in next section?

Dockerfile
```diff
@@ -0,0 +1,34 @@
+# Pull base image
+FROM --platform=linux/x86_64 continuumio/miniconda3:23.10.0-1
+
+# Add files
+ADD ./tutorials /home/deeprank2/tutorials 
+ADD ./env/environment.yml /home
```
dsmits, line 6: It doesn't make sense to add it in `/home`. `/home` is the directory in linux that contains all the users' home directories. You don't save other data in it. I would put it in /home/deeprank2 where the tutorials are as well. If you want to hide it from the user I would add it in a totally different dir, like `/opt/deeprank2`, or just `/deeprank2`.

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+ADD ./env/requirements.txt /home
```
dsmits, line 7: Same comment as for the `environment.yml` file.

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+# Activate the environment
+RUN echo "source activate deeprank2" > ~/.bashrc
+ENV PATH /opt/conda/envs/deeprank2/bin:$PATH
+
+# Define working directory
+WORKDIR /home/deeprank2
+
+# Define default command
+CMD ["bash"]
```
dsmits, line 34: Don't you want to start up a jupyter server instead?

README.md
```diff
@@ -37,42 +37,81 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
+
+```bash
+docker run -it --expose 3000  -p 3000:3000 deeprank2
+```
+
+Run the tutorials' notebooks from within the running container:
+```bash
+cd tutorials
+jupyter notebook --ip 0.0.0.0 --no-browser --allow-root --port 3000
+```
```
DaniBodor, line 76: I needed to run `sudo` before any of that. I guess it depends on permissions set up (@dsmits sent me a link to make docker not depend on sudo anymore), but maybe better to just include it in the command blocks, as I don't think it hurts.

I can't add the suggestion, as the "suggestion block" gets killed by the triple-` from the bash code block :)

README.md
```diff
@@ -37,42 +37,81 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
 
-Before installing deeprank2 you need to install some dependencies. We advise to use a [conda environment](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html) with Python >= 3.10 installed. The following dependency installation instructions are updated as of 14/09/2023, but in case of issues during installation always refer to the official documentation which is linked below:
+In order to try out the package without worrying about your OS and without the need of installing all the required dependencies, we created a `Dockerfile` that can be used for taking care of everything in a suitable container. After having cloned the repository and installed [Docker](https://docs.docker.com/engine/install/), run the following commands from the root of the repository.
+
+Build the Docker image:
+```bash
+docker build -t deeprank2 .
+```
+
+SSH to a running container:
```
dsmits, line 66:
```suggestion
Start a container in interactive mode:
```
Although I would actually just replace the CMD in the dockerfile with `jupyter notebook` etc so you can skip this step.

README.md
```diff
@@ -37,42 +37,81 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
+```
+
+Run the tutorials' notebooks from within the running container:
+```bash
+cd tutorials
+jupyter notebook --ip 0.0.0.0 --no-browser --allow-root --port 3000
+```
+
+Now you can run the tutorials' notebook. More details about their content can be found [here](https://github.com/DeepRank/deeprank2/blob/main/tutorials/TUTORIAL.md). Note that in the docker container only the raw PDB files are downloaded, needed as a starting point for the tutorials. You can obtain the processed HDF5 files by running the `data_generation_xxx.ipynb` notebooks. Because Docker containers are limited in memory resources, we limit the number of data points processed in the tutorials'. Please install the package locally to fully leverage its capabilities.
+
```
DaniBodor, line 79: When a user has finished testing it, should they then remove the docker image? Maybe also add instructions on how to do that, as it is quite a large thing to have stored on HD if not used anymore.

Dockerfile
dsmits, line 34:
```suggestion
CMD ["/opt/conda/envs/deeprank2/bin/jupyter", "notebook"]
```

README.md
```diff
@@ -37,42 +37,81 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
+Build the Docker image:
+```bash
+docker build -t deeprank2 .
+```
+
+SSH to a running container:
+
+```bash
+docker run -it --expose 3000  -p 3000:3000 deeprank2
```
dsmits, line 69: If you change the CMD then this instruction needs to be changed to:

```suggestion
docker run  -p 8888:8888 deeprank2
```

8888 is the default port for jupyter

README.md
```diff
@@ -37,42 +37,81 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
+```
+
+Run the tutorials' notebooks from within the running container:
```
dsmits, line 72: If you change the CMD these instructions aren't necessary anymore

README.md
```diff
@@ -37,42 +37,81 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
+jupyter notebook --ip 0.0.0.0 --no-browser --allow-root --port 3000
+```
+
+Now you can run the tutorials' notebook. More details about their content can be found [here](https://github.com/DeepRank/deeprank2/blob/main/tutorials/TUTORIAL.md). Note that in the docker container only the raw PDB files are downloaded, needed as a starting point for the tutorials. You can obtain the processed HDF5 files by running the `data_generation_xxx.ipynb` notebooks. Because Docker containers are limited in memory resources, we limit the number of data points processed in the tutorials'. Please install the package locally to fully leverage its capabilities.
+
+### Non-pythonic dependencies
+
+Instructions are updated as of 14/09/2023.
+
+Before installing deeprank2 you need to install some dependencies:
```
DaniBodor, line 84: I think "Non-pythonic dependencies" and "Pythonic dependencies" should be one tick-level lower than "Dockerfile", as they together form the "local installation" (or whatever we want to call that, maybe "full installation"?).

```suggestion
### Local installation
#### Non-pythonic dependencies

Instructions are updated as of 14/09/2023.

Before installing deeprank2 you need to install some dependencies:
```

README.md
```diff
@@ -37,42 +37,81 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
+```bash
+docker run -it --expose 3000  -p 3000:3000 deeprank2
+```
+
+Run the tutorials' notebooks from within the running container:
+```bash
+cd tutorials
+jupyter notebook --ip 0.0.0.0 --no-browser --allow-root --port 3000
+```
+
```
dsmits, line 77: Say something like:
Open a browser and go to `localhost:3000` (or localhost:8888 if you use my suggestions)

docs/installation.md
```diff
@@ -1,28 +1,65 @@
 # Installation
 
-The package officially supports ubuntu-latest OS only, whose functioning is widely tested through the continuous integration workflows. 
+Note that the package officially supports ubuntu-latest OS only, whose functioning is widely tested through the continuous integration workflows. 
 
-## Dependencies
+## Dockerfile
```
dsmits, line 5: Same remarks as for README.md

tutorials/data_generation_ppi.ipynb
dsmits, line 1: Don't know if `limit_data` is necessary. Docker containers usually don't have resource constraints. They have access to the same amount of memory as the host system

dsmits (changes requested):
Very nice, couple of remarks

README.md
```diff
@@ -37,42 +37,81 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
-### Dependencies
+### Dockerfile
 
-Before installing deeprank2 you need to install some dependencies. We advise to use a [conda environment](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html) with Python >= 3.10 installed. The following dependency installation instructions are updated as of 14/09/2023, but in case of issues during installation always refer to the official documentation which is linked below:
+In order to try out the package without worrying about your OS and without the need of installing all the required dependencies, we created a `Dockerfile` that can be used for taking care of everything in a suitable container. After having cloned the repository and installed [Docker](https://docs.docker.com/engine/install/), run the following commands from the root of the repository.
+
+Build the Docker image:
+```bash
+docker build -t deeprank2 .
+```
```
DaniBodor, line 64: Did not work for me:
> ERROR: failed to solve: process "/bin/sh -c apt update -y &&   apt install unzip -y &&   apt install -y gcc &&   wget https://github.com/PDB-REDO/dssp/releases/download/v4.4.0/mkdssp-4.4.0-linux-x64 &&   mv mkdssp-4.4.0-linux-x64 /usr/local/bin/mkdssp &&   chmod a+x /usr/local/bin/mkdssp &&   conda env create -f /home/environment.yml &&   wget https://zenodo.org/records/8349335/files/data_raw.zip &&   unzip data_raw.zip -d data_raw &&   mv data_raw /home/deeprank2/tutorials" did not complete successfully: exit code: 1

Scrolling further up, I noticed:
> 168.0 mv: cannot move 'data_raw' to '/home/deeprank2/tutorials/data_raw': Directory not empty

After discussing with @dsmits, we realized that the problem is that I had already downloaded the raw data into my local folder (with same name) and that that is why the docker installation cannot add the same files. Another thing I noticed is that because I also have the "processed" data folder in my local repo, the docker image will also use those, and that is multiple GB of data, which we don't necessarily want/need in the docker image.

In any case, there is some conflict between having downloaded the data already and docker trying to download it that needs to be resolved.

README.md
```diff
@@ -37,42 +37,81 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
 *  [DSSP 4](https://swift.cmbi.umcn.nl/gv/dssp/)
     * Check if `dssp` is installed: `dssp --version`. If this gives an error or shows a version lower than 4:
       * on ubuntu 22.04 or newer: `sudo apt-get install dssp`. If the package cannot be located, first run `sudo apt-get update`.
       * on older versions of ubuntu or on mac or lacking sudo priviliges: install from [here](https://github.com/pdb-redo/dssp), following the instructions listed. Alternatively, follow [this](https://github.com/PDB-REDO/libcifpp/issues/49) thread. 
 *  [GCC](https://gcc.gnu.org/install/)
-    * Check if gcc is installed: `gcc --version`. If this gives an error, run `sudo apt-get install gcc`.  
-*  For MacOS with M1 chip users only install [the conda version of PyTables](https://www.pytables.org/usersguide/installation.html).
+    * Check if gcc is installed: `gcc --version`. If this gives an error, run `sudo apt-get install gcc`. 
+
+### Pythonic dependencies
```
DaniBodor, line 93: see comment above
```suggestion
#### Pythonic dependencies
```

README.md
```diff
@@ -37,42 +37,81 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
   - [Computational performances](#computational-performances)
   - [Package development](#package-development)
 
 ## Installation
 
-The package officially supports ubuntu-latest OS only, whose functioning is widely tested through the continuous integration workflows. 
+Note that the package officially supports ubuntu-latest OS only, whose functioning is widely tested through the continuous integration workflows. 
 
-### Dependencies
+### Dockerfile
```
DaniBodor, line 57: Maybe more generic description

```suggestion
### Containerized installation
```

DaniBodor (changes requested):
I'm not done reviewing, but will already release these comments, as we will need some changes already. I will continue from here with a new review

README.md
```diff
@@ -37,42 +37,81 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
+
+*  [MSMS](https://anaconda.org/bioconda/msms): `conda install -c bioconda msms`.
+    * [Here](https://ssbio.readthedocs.io/en/latest/instructions/msms.html) for MacOS with M1 chip users.
+*  [PyTorch](https://pytorch.org/get-started/locally/)
+*  [PyTorch Geometric](https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html) and its optional dependencies: `torch_scatter`, `torch_sparse`, `torch_cluster`, `torch_spline_conv`.
+*  For MacOS with M1 chip users only install [the conda version of PyTables](https://www.pytables.org/usersguide/installation.html).
 
-Once the dependencies are installed, you can install the latest stable release of deeprank2 using the PyPi package manager:
+Finally do:
```
DaniBodor, line 114:
```suggestion
#### Install DeepRank2

Finally do:
```

in the section below (I can't comment there, currently line 120, previously 81), I would state "get the latest updates" instead of "get all the new developments".

DaniBodor (changes requested):
Here's the rest of my review.

I also tested installating the env from yml file, as that is new too, and works fine (ran all tests and they passed)

tutorials/data_generation_ppi.ipynb
gcroci2: It does for me. For running the preprocessing for 100 data points only it was taking more than 10 mins, operation that on my Mac takes a couple of mins. Maybe it's because of the local host? Or since we use multiprocessing, for some reasons its functionality is limited in a container? @dsmits

README.md
gcroci2: For me, running everything on the docker container was much much slower. I've asked Djura why this could be, I was imagining that for reasons the resources available in a container are not the same that you have on your machine.

README.md
gcroci2: On the other hand, many users would use a super computer like snellius and wouldn't have sudo permissions. I think that this is beyond the scope of these instructions anyway.

README.md
gcroci2: If you follow the instructions, it is written that you should clone the repo from scratch and then build the image (so no data will be in tutorials/). Also, having downloaded the data already is not something that a user would have when trying out the package for the first time.

tutorials/data_generation_ppi.ipynb
DaniBodor: This reminds me of [this](https://github.com/DeepRank/deeprank2/issues/494#issuecomment-1708559180) comment I made a while ago in an open issue. it's a bit of scope creep, but maybe adjust that here as well?
> we should decrease the distance_cutoff and radius in both atomic level tutorials to 3.0, otherwise it takes way too long to generate the data (took >25 mins for me and then I stopped it).

README.md
DaniBodor: Good point!
Maybe at least mention that a user might need sudo permission to run. I think for many inexperienced users (which would usually want to just run the docker), it would be unclear why they get a permission error when trying to build the container if just running those lines.

README.md
DaniBodor: We should add a comment about when a user should/would want to install it in which of the two ways. Maybe local installation would really only be for developers, if the docker turns out to be equally fast, etc.

README.md
DaniBodor: I still feel that it's confusing and error prone. If someone doesn't follow the steps to the letter (maybe already started looking at the tutorials) and downloads the data beforehand or if (like me), their build will fail. 
Either we should be more explicit about making sure those folders are empty, or create a workaround for this scenario. Personally, I don't think it should have to be required to re-clone the repo. If anyone already has the repo, the assumption will be that by just pulling the latest updates, the build should work and it's kind of annoying to have to clone a fresh one.

README.md
DaniBodor: I don't really agree with this. I think, especially if we want to make this a widely used tool, the installation instructions should be as dummy proof as possible. We don't need detailed instructions for this or above (and below), but it doesn't hurt to have a few words on this.

Dockerfile
gcroci2: Indeed I meant to hide them from the user. I edited `/home` with `/deeprank`

tutorials/data_generation_ppi.ipynb
gcroci2: I did it :) now it's 3.0 everywhere. But my concern above remains.

Dockerfile
gcroci2: Changed now

gcroci2: This wasn't working. I used the following instead:
`CMD ["jupyter", "notebook", "--ip=0.0.0.0", "--NotebookApp.token=''","--NotebookApp.password=''", "--allow-root"]`

README.md
gcroci2: For real use cases, when a user needs to process thousands of data, I don't see any easy way of making use of the docker container (they will need to submit jobs to the machines of a supercomputer cluster, as we did in Snellius). For now, I'd just stick to the advice to use the docker option for trying out the package, otherwise the "normal" installation. If and when we get a wider user base, we can always improve the instructions and investigate the options.

OT: that's also the reason why I don't want to call the "normal" installation section "local" (since it's not always the case, for example, if you want to install the package via SSH).

README.md
gcroci2: I added some general info and refs about that

README.md
gcroci2: It should be fixed now

tutorials/data_generation_ppi.ipynb
```diff
@@ -102,7 +102,9 @@
     "data_path = os.path.join(\"data_raw\", \"ppi\")\n",
     "processed_data_path = os.path.join(\"data_processed\", \"ppi\")\n",
     "os.makedirs(os.path.join(processed_data_path, \"residue\"))\n",
-    "os.makedirs(os.path.join(processed_data_path, \"atomic\"))"
+    "os.makedirs(os.path.join(processed_data_path, \"atomic\"))\n",
```
DaniBodor, line 105: By setting `exist_ok` to `True` (also needs to be done above and in other notebook), the notebook can be run a second time without having to remove the folder created during previous run.

```suggestion
    "os.makedirs(os.path.join(processed_data_path, \"atomic\", exist_ok=True))\n",
```

tutorials/data_generation_ppi.ipynb
```diff
@@ -139,7 +141,10 @@
     "\tbas = csv_data_indexed.measurement_value.values.tolist()\n",
     "\treturn pdb_files, bas\n",
     "\n",
-    "pdb_files, bas = get_pdb_files_and_target_data(data_path)"
+    "pdb_files, bas = get_pdb_files_and_target_data(data_path)\n",
+    "\n",
+    "if limit_data:\n",
+    "\tpdb_files = pdb_files[:10]"
```
DaniBodor, line 147: with 10, the training crashes because the test set is too small or sth. Ok at 15 though
```suggestion
    "\tpdb_files = pdb_files[:15]"
```

tutorials/data_generation_srv.ipynb
```diff
@@ -158,7 +160,10 @@
     "\tpdb_files = [data_path + \"/pdb/\" + pdb_name for pdb_name in pdb_names]\n",
     "\treturn pdb_files, res_numbers, res_wildtypes, res_variants, targets\n",
     "\n",
-    "pdb_files, res_numbers, res_wildtypes, res_variants, targets = get_pdb_files_and_target_data(data_path)"
+    "pdb_files, res_numbers, res_wildtypes, res_variants, targets = get_pdb_files_and_target_data(data_path)\n",
+    "\n",
+    "if limit_data:\n",
+    "\tpdb_files = pdb_files[:10]"
```
DaniBodor, line 166: see above
```suggestion
    "\tpdb_files = pdb_files[:15]"
```

DaniBodor (changes requested):
I'm getting an error in the trainer notebook if I run it now (in the cell where GNN is tested). I get the same error if I run it locally from this branch and the latest env, but not from my working repo/env.


> 
> UnboundLocalError                         Traceback (most recent call last)
> Cell In[9], line 7
>       4 earlystop_maxgap = 0.1
>       5 min_epoch = 10
> ----> 7 trainer.train(
>       8     nepoch = epochs,
>       9     batch_size = batch_size,
>      10     earlystop_patience = earlystop_patience,
>      11     earlystop_maxgap = earlystop_maxgap,
>      12     min_epoch = min_epoch,
>      13     validate = True,
>      14     filename = os.path.join(output_path, f"gnn_{task}", "model.pth.tar"))
>      16 epoch = trainer.epoch_saved_model
>      17 print(f"Model saved at epoch {epoch}")
> 
> File /opt/conda/envs/deeprank2/lib/python3.10/site-packages/deeprank2/trainer.py:640, in Trainer.train(self, nepoch, batch_size, shuffle, earlystop_patience, earlystop_maxgap, min_epoch, validate, num_workers, best_model, filename)
>     638 # Now that the training loop is over, save the model
>     639 if filename:
> --> 640     torch.save(checkpoint_model, filename)
>     641 self.opt_loaded_state_dict = checkpoint_model["optimizer_state"]
>     642 self.model_load_state_dict = checkpoint_model["model_state"]
> 
> UnboundLocalError: local variable 'checkpoint_model' referenced before assignment

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+  apt install -y gcc && \
+  ## DSSP
+  wget https://github.com/PDB-REDO/dssp/releases/download/v4.4.0/mkdssp-4.4.0-linux-x64 && \
+  mv mkdssp-4.4.0-linux-x64 /usr/local/bin/mkdssp && \
+  chmod a+x /usr/local/bin/mkdssp && \
+  ## Conda and pip deps
+  conda env create -f /home/deeprank2/environment.yml && \
+  ## Get the data for running the tutorials
+  if [ -d "/home/deeprank2/tutorials/data_raw" ]; then rm -Rf /home/deeprank2/tutorials/data_raw; fi && \
+  if [ -d "/home/deeprank2/tutorials/data_processed" ]; then rm -Rf /home/deeprank2/tutorials/data_processed; fi && \
```
DaniBodor, line 23: this only deletes the folders from the image, not the existing repo, right?

Dockerfile
```diff
@@ -0,0 +1,36 @@
+# Pull base image
+FROM --platform=linux/x86_64 continuumio/miniconda3:23.10.0-1
```
jnwei, line 2: As a small suggestion, you might see if mambaforge can support your build, as it can make the environment build a lot faster. https://hub.docker.com/r/condaforge/miniforge3

tutorials/data_generation_ppi.ipynb
gcroci2: But if you've already processed data, it will keep adding hdf5 files to the same folder, even if during the second run you may have edited the settings of the processing. I think that the user can edit by himself/herself the code in such cases if he/she means to do that. We're providing the notebooks to show them a working pipeline, we don't have to consider all the possibilities in this regard.

README.md
```diff
@@ -36,49 +36,98 @@ DeepRank2 extensive documentation can be found [here](https://deeprank2.rtfd.io/
...
+
+After running the tutorials, you may want to remove the (quite large) Docker image from your machine. In this case, remember to [stop the container](https://docs.docker.com/engine/reference/commandline/stop/) and then [remove the image](https://docs.docker.com/engine/reference/commandline/image_rm/). More general information about Docker can be found on the [official website docs](https://docs.docker.com/get-started/). 
+
+### Local/remote installation
+
+#### Non-pythonic dependencies
+
+Instructions are up to date as of 14/09/2023.
+
+Before installing deeprank2 you need to install some dependencies:
```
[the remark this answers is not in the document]
gcroci2: When we refer to the PyPi installation or to the package as a folder in GitHub, then we use deeprank2. Otherwise DeepRank2. I uniformed this everywhere.

Dockerfile
gcroci2: exactly

gcroci2:
> I'm getting an error in the trainer notebook if I run it now (in the cell where GNN is tested). I get the same error if I run it locally from this branch and the latest env, but not from my working repo/env.
> 
> > UnboundLocalError                         Traceback (most recent call last)
> > Cell In[9], line 7
> > 4 earlystop_maxgap = 0.1
> > 5 min_epoch = 10
> > ----> 7 trainer.train(
> > 8     nepoch = epochs,
> > 9     batch_size = batch_size,
> > 10     earlystop_patience = earlystop_patience,
> > 11     earlystop_maxgap = earlystop_maxgap,
> > 12     min_epoch = min_epoch,
> > 13     validate = True,
> > 14     filename = os.path.join(output_path, f"gnn_{task}", "model.pth.tar"))
> > 16 epoch = trainer.epoch_saved_model
> > 17 print(f"Model saved at epoch {epoch}")
> > File /opt/conda/envs/deeprank2/lib/python3.10/site-packages/deeprank2/trainer.py:640, in Trainer.train(self, nepoch, batch_size, shuffle, earlystop_patience, earlystop_maxgap, min_epoch, validate, num_workers, best_model, filename)
> > 638 # Now that the training loop is over, save the model
> > 639 if filename:
> > --> 640     torch.save(checkpoint_model, filename)
> > 641 self.opt_loaded_state_dict = checkpoint_model["optimizer_state"]
> > 642 self.model_load_state_dict = checkpoint_model["model_state"]
> > UnboundLocalError: local variable 'checkpoint_model' referenced before assignment

The issue is that if the loss is nan (such as in this case using very few datapoints), `checkpoint_model` is never created, thus the error. Will be fixed in #533

README.md
gcroci2: I opened an issue about this discussion (#534 )

tutorials/data_generation_ppi.ipynb
gcroci2: I realized that my Docker installation wasn't using all my machine's resources. I had to modify the Docker's advanced settings manually, and still not all the CPUs can be used (8 out of 10) and not all the RAM. Reading online, I learned that Docker may not have access to all the system resources.

gcroci2: Again, this discussion will go on in #534; I will close this PR asap in the meantime, for the sake of the JOSS review.

Dockerfile
gcroci2: I will merge soon the PR, but will leave this conversation unsolved so you can comment when you come back :) @dsmits

PR #554, opened 2024-01-15
ci: implement ruff

### LINKED ISSUE #541: Change linter to ruff

DaniBodor:
Implement ruff linter and formatter in deeprank in lieu of prospector.

best reviewed one commit at a time:
1. set up ruff
    - replace prospector by ruff in CI
    - add ruff settings to `pyproject.toml` and remove `.prospector.yml`
    - format `pyproj.toml`
2. update docs regarding new linting/formatting
3. update settings.json for vs code
    - automatically flag linting issues and apply formatting upon save of python files
    - some additional settings
4. implement new linting and formatting throughout
    - this is a huge commit, basically touching the entire code base; probably not worthwhile reviewing in detail. In short
        - auto-format the entire code base (using `ruff format .`)
        - manually or automatically fix all linting issues flagged by `ruff .` (or suppressed)
        - remove pylint suppression calls
        - additional manual formatting to error messages to avoid them printing lots of whitespace/tabs when raised
5. remove all commented out code (linting issue ERA001), this part should be reviewed.
    - this was done in a separate commit to check that all deleted code is indeed legacy as opposed to accidentally being commented out
    - some commented out code may contain information that should be stored in a more appropriate manner
6. suppress some warnings during testing
    - I would like to get to the bottom of the warnings thrown during test_trainer, but that took too much time and will be for another issue.
7. format non-python files
    - using [prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) for most file formats.

closes #541


README.dev.md
```diff
@@ -42,9 +42,9 @@ coverage report
 
 `coverage` can also generate output in HTML and other formats; see `coverage help` for more information.
 
-## Linting
+## Linting and Formatting
 
-We use [prospector](https://pypi.org/project/prospector/) with pyroma for linting. For running it locally, use `prospector` or `prospector <filepath_or_folderpath>` for specific files/folders.
+We use [ruff](https://docs.astral.sh/ruff/) for linting, sorting imports and formatting code. The configurations of `ruff` are set in [pyproject.toml](pyproject.toml) file. If you are using VS code, you can activate the [Ruff extension](charliermarsh.ruff) to automatically format the code and flag linting issues upon saving. Otherwise, please check both linting (`ruff .`) and formatting (`ruff format .`, this automatically formats the entire code base) before requesting a review.
```
gcroci2, line 47: Add the correct link for project.toml file

tests/__init__.py
gcroci2, line 1: I guess that ruff added it. Do we know why? Is also added init files in each tests subfolder

gcroci2 (approved):
Nice work :D I left very minor comments and seems all great. Three other things:
- You used [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) on non-python files, what about adding it as the default formatter for non-python files in the vscode settings? It would be even better to have ruff as the default for non-python files as well, but I think that ruff is for python files only, right?
- I had a lower version of ruff (0.1.4 instead of the latest, 0.1.13) and it was giving me 4 errors when running `ruff .` from local. No errors with the updated ruff version. I'd add the version on the dependencies list in the toml file (ruff >= 0.1.13)
- When I run `ruff format .`, I get the following warning:
```bash
The following rules may cause conflicts when used with the formatter: 'ISC001'. To avoid unexpected behavior, we recommend disabling these rules, either by removing them from the 'select' or 'extend-select' configuration, or adding them to the 'ignore' configuration.
86 files left unchanged
```
Maybe is a good idea to disable [ISC001](https://docs.astral.sh/ruff/rules/single-line-implicit-string-concatenation/) adding it to the ignore configuration in the toml file?

DaniBodor:
> Nice work :D I left very minor comments and seems all great. Three other things:
> 
>   * You used [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) on non-python files, what about adding it as the default formatter for non-python files in the vscode settings? It would be even better to have ruff as the default for non-python files as well, but I think that ruff is for python files only, right?

Done in 7dc25266 (edited _style: format non-python files_)

> 
>   * I had a lower version of ruff (0.1.4 instead of the latest, 0.1.13) and it was giving me 4 errors when running `ruff .` from local. No errors with the updated ruff version. I'd add the version on the dependencies list in the toml file (ruff >= 0.1.13)

Done in b5767f6b (edited _ci: set up ruff linter_)

> 
>   * When I run `ruff format .`, I get the following warning:
> 
> 
> ```shell
> The following rules may cause conflicts when used with the formatter: 'ISC001'. To avoid unexpected behavior, we recommend disabling these rules, either by removing them from the 'select' or 'extend-select' configuration, or adding them to the 'ignore' configuration.
> 86 files left unchanged
> ```
> 
> Maybe is a good idea to disable [ISC001](https://docs.astral.sh/ruff/rules/single-line-implicit-string-concatenation/) adding it to the ignore configuration in the toml file?

This is because the format may put strings that are on sequential lines onto the same line but keeps them as two separate strings and this creates the linting issue you reference. I think we do want the formatter to do this and for the linter to pick it up.
I now added that linting issue to the "safe to fix" list (in b5767f6b), so that it automatically get fixed upon saving as well. Stupidly however, you need to save twice (first save will move the two strings onto one line, the second save will merge into a single string) for the full issue to resolve. Also, the warning persists and I haven't found a way to suppress it.

README.dev.md
DaniBodor: The link works like this.

tests/__init__.py
DaniBodor: from [INP001](https://docs.astral.sh/ruff/rules/implicit-namespace-package/):

> Python packages are directories that contain a file named __init__.py. The existence of this file indicates that the directory is a Python package, and so it can be imported the same way a module can be imported.
> 
> Directories that lack an __init__.py file can still be imported, but they're indicative of a special kind of package, known as a "namespace package" (see: [PEP 420](https://www.python.org/dev/peps/pep-0420/)). Namespace packages are less widely used, so a package that lacks an __init__.py file is typically meant to be a regular package, and the absence of the __init__.py file is probably an oversight.

PR #563, opened 2024-01-31
Adapting ruff

DaniBodor:
Activated and fixed more settings.

@gcroci2 , I think most of these you can just glance over, but can you please check what I did in 35248b9. This was where I dealt with print statements in the code base, either by:
- sending to log instead
- generating a warning
- left as is (suppress the lint rule)
- removed the print statement with no replacement.
Can you check whether you agree with what I did where.


gcroci2 (approved):
Nice edits, now the toml file is much shorter and the code seems nicer and nicer :D I went through commit [35248b9](https://github.com/DeepRank/deeprank2/commit/35248b9c0333a4c185ed32f472fac632166bb7d1) and all seems good to me

PR #572, opened 2024-02-21
fix: test_cuda in `test_trainer.py::TestTrainer`

gcroci2:
In `tests/test_trainer.py::TestTrainer::test_cuda`, we were not setting cuda correctly - not using the `use_cuda` parameter in the test.
We were also asserting that data tensors are cuda tensors, but this is never true since the data are transferred to the GPU device only during the training when iterating over the Dataloader instances ([see line 705](https://github.com/DeepRank/deeprank2/blob/822927025241f20230b0a18444246b755746de0e/deeprank2/trainer.py#L705) of trainer.py).


DaniBodor (approved):
nice catch

PR #573, opened 2024-02-21
fix: features transform in `Trainer._save_model` internal method

gcroci2:
- The bug was due to the quotation marks type used in the regex for extracting the transform function. The quotation marks are indeed now automatically fixed to double type by ruff in our VSCode editor. Since a user should be able to use both types of quotation marks, I changed the regex that matches the `transform` function for handling both cases. The bug was causing the failure of the `training.ipynb` notebook.
- I also added a test for checking on the `features_transform` attribute in the `Trainer`, and in case we use a pre-trained model. The testing on the trainer side for features_transform was missing. 

As soon as this is approved, I will release a new patch version for deeprank2.

### DIFF
--- a/deeprank2/trainer.py
+++ b/deeprank2/trainer.py
@@ -947,7 +947,7 @@ class Trainer:
                 if key["transform"] is None:
                     continue
                 str_expr = inspect.getsource(key["transform"])
-                match = re.search(r"\'transform\':.*(lambda.*).*,.*\'standardize\'.*", str_expr).group(1)
+                match = re.search(r"[\"|\']transform[\"|\']:.*(lambda.*).*,.*[\"|\']standardize[\"|\'].*", str_expr).group(1)
                 key["transform"] = match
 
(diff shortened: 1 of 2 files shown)


DaniBodor (approved):
Nice catch, once again.

If ever there is some extra time on this project (not likely), it would be nice to refactor the entire tests folder. Many of these modules are getting really long and confusing and contain lots of duplicate code. It's hard to keep track of what is going on. A nice cleanup could make it really easy to quickly assess what each test does and how.

gcroci2:
> Nice catch, once again.
> 
> If ever there is some extra time on this project (not likely), it would be nice to refactor the entire tests folder. Many of these modules are getting really long and confusing and contain lots of duplicate code. It's hard to keep track of what is going on. A nice cleanup could make it really easy to quickly assess what each test does and how.

Indeed, it would be very nice. We also have already issues for that. Let's see what budgets we get ;)

PR #587, opened 2024-03-04
ci: add action for testing the tutorials' notebooks

gcroci2:


.github/workflows/notebooks.yml
```diff
@@ -0,0 +1,64 @@
...
+on:
+  push:
+    paths-ignore:
+      # specific folder locations
+      - ".vscode/**"
+      - "docs/**"
+      # filetypes
+      - "**.md"
+      - "**.rst"
+      - "**.ipynb"
```
DaniBodor, line 12: I don't think we want to ignore these in this case
```suggestion
```

.github/workflows/notebooks.yml
```diff
@@ -0,0 +1,64 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths-ignore:
+      # specific folder locations
+      - ".vscode/**"
+      - "docs/**"
+      # filetypes
+      - "**.md"
+      - "**.rst"
+      - "**.ipynb"
```
DaniBodor, line 26:
```suggestion
```

.github/workflows/notebooks.yml
gcroci2: Definitely!! Good catch :)

PR #612, opened 2024-06-18
ci: simplify lint action

DaniBodor:
Linting doesn't require the entire package to be built, merely ruff to be installed.

Note that the lint action only takes 7 sec now, compared to e.g. 5 min in #611


gcroci2 (approved):
Nice :D

Linting is now failing, because of changes in ruff I guess. Pls fix that and feel free to merge

PR #615, opened 2024-06-20
build: simplify docker installation and improve the manual one

gcroci2:


DaniBodor (approved):
So, I am looking at the install instructions now.
We are basically asking people to clone the entire (fairly large) repo only to access the environment yml file and then pip install it from elsewhere. Given that we anyway have the installation commands copy/pastable, would it perhaps make sense to have a command to download just that one file instead? Or is it even possible to create the env from the online file?

If it's complicated then I think it's fine to just leave it though.

gcroci2:
> So, I am looking at the install instructions now. We are basically asking people to clone the entire (fairly large) repo only to access the environment yml file and then pip install it from elsewhere. Given that we anyway have the installation commands copy/pastable, would it perhaps make sense to have a command to download just that one file instead? Or is it even possible to create the env from the online file?
> 
> Not sure if it's worthwhile changing, just noticed this and thought I'd mention it :)

Very good point! I changed them in https://github.com/DeepRank/deeprank2/pull/615/commits/d424a26d62dc21513297ac9ae2306fba4b4d7bfd, let me know if you like it better now :)

DaniBodor:
> > So, I am looking at the install instructions now. We are basically asking people to clone the entire (fairly large) repo only to access the environment yml file and then pip install it from elsewhere. Given that we anyway have the installation commands copy/pastable, would it perhaps make sense to have a command to download just that one file instead? Or is it even possible to create the env from the online file?
> > Not sure if it's worthwhile changing, just noticed this and thought I'd mention it :)
> 
> Very good point! I changed them in [d424a26](https://github.com/DeepRank/deeprank2/commit/d424a26d62dc21513297ac9ae2306fba4b4d7bfd), let me know if you like it better now :)

Nice! 
I guess the other thing is that if it wasn't cloned, the tests won't work, so the testing part should maybe be deleted or adjusted as well?

DaniBodor (approved):

gcroci2:
> Nice! I guess the other thing is that if it wasn't cloned, the tests won't work, so the testing part should maybe be deleted or adjusted as well?

I removed the ref in that part, thanks :)

> Also, requirements-docker.txt can be removed from the bumpversion config file.

Good catch, removed!

PR #620, opened 2024-07-01
docs: clarify tutorials usage

gcroci2:
I added a disclaimer in the tutorials' readme to make user that users know that they are not supposed to run the notebooks' code into python interactive shell.


tutorials/TUTORIAL.md
```diff
@@ -7,6 +7,10 @@ The tutorial notebooks in this folder can be run to learn how to use DeepRank2.
   - single-residue variant (SRV) data ([data_generation_srv.ipynb](https://github.com/DeepRank/deeprank2/blob/main/tutorials/data_generation_srv.ipynb)).
 - The [training tutorial](tutorials/training_ppi.ipynb) will demonstrate how to train neural networks using DeepRank2.
 
+### :heavy_exclamation_mark: Important :heavy_exclamation_mark:
```
DaniBodor, line 10: Not sure this level of emphasis is needed :)

tutorials/TUTORIAL.md
```diff
@@ -7,6 +7,10 @@ The tutorial notebooks in this folder can be run to learn how to use DeepRank2.
...
+
+For correct results, please run these tutorials within the expected Jupyter notebooks (`.ipynb` files) and avoid using the Python interactive shell (command line), which may cause certain functions (e.g., such as `inspect.getsource()`) to not work as expected.
```
DaniBodor, line 12:
```suggestion
For a smooth experience, please run these tutorials within the Jupyter notebooks provided (`.ipynb` files). Running the commands in a Python (interactive) shell or on command line may lead to unexpected errors.
```

DaniBodor (approved):
Good addition. This clarifies that users should use the system we provide.
I do think we can be less explicit about what goes wrong if they don't (as per my suggestion below).

Also, should this PR point to dev instead of main? or is that not needed if it's just docs changes? It looks like it was branched off of dev, but now points to main.

gcroci2:
> Good addition. This clarifies that users should use the system we provide. I do think we can be less explicit about what goes wrong if they don't (as per my suggestion below).

I accepted your suggestion and I removed the section's title (thinking about it the note is enough)

> Also, should this PR point to dev instead of main? or is that not needed if it's just docs changes? It looks like it was branched off of dev, but now points to main.

Indeed, moved to dev :)

Issue #608, opened 2024-06-12
Docs - create descriptions/uses for the neural networks


rubenalv:
For the naive user, would it be possible to add a small description and preferred application(s) of the different neural networks supplied?

DaniBodor:
Excellent idea!
Thanks for suggesting this/picking up on the fact that we don't have that yet 😄

PR #625, opened 2024-07-09
docs: improve structure and add modules

gcroci2:
You can look at the new docs rendered [here](https://deeprank2.readthedocs.io/en/297_increase_modules_docs_gcroci2/index.html).

I suggest to not look at the commits for this PR, but directly to the files changes. Main changes:
- The modules rendered are now dataset, query, trainer, neuralnets, molstructu, tools (before only dataset was successfully rendered, see [here](https://deeprank2.readthedocs.io/en/latest/reference/deeprank2.html)). I removed features since we didn't add any docs there, so it's useless to render those submodules on readthedocs. 
- I fixed various cross sections refs and links in the installation page. 
- Code of conduct and contributing files are now .md files, a preferred file format (easier to handle).
- Contributing page in the docs is now linked to the one we have in the repo's main root (less duplication). 
- Uniform query.py to dataset.py in terms of doc strings (both have a parent class which is used by two children classes; the args are explained in detail only in the children, not in the parents - otherwise it's visually very verbose).
- Add `from __future__ import annotations` for making possible the rendering of modules that were not correctly rendered before (or not rendered at all). To fix the issue, also `TYPE_CHECKING` is used. Briefly, there were problems with type hints that made sphinx build fail on some modules (an error in the type hints import caused the modules to not be rendered). 
- Now the docs/ folder has a source/ subfolder. This is useful for building the docs locally (where you can have a build/ folder as well, not uploaded to the remote of course).

Note for myself:
- [x] Hide the branch on readthedocs when the PR is merged.


DaniBodor (changes requested):
Great updates, the docs are definitely much more complete now.
Were you planning on adding descriptions for the NNs (as per #608) in this PR or in a separate one?

I added a commit that adds empty lines at the end of a couple of files were that wasn't the case

DaniBodor (approved):

gcroci2:
> Great updates, the docs are definitely much more complete now. Were you planning on adding descriptions for the NNs (as per #608) in this PR or in a separate one?

Separate PR, but I've already added the neural networks submodule so it's ready for issue #608 :)

PR #626, opened 2024-07-09
style: make torch.nn imports more uniform

### LINKED ISSUE #269: Unify torch imports
Unify import torch.nn.Module vs torch.nn vs torch in neuralnet modules (and other places)

DaniBodor:
- All functions from torch.nn.functional are now explicitly imported by name and used directly.
- In all other cases where classes from torch.nn are used, we do `from torch import nn` and then use `nn.<Class>` throughout.

Note that this was also changed in one docstring in fba9e56. Not sure if that messes up the rendering of the documents. That commit can easily be dropped if needed.

fix #269

### DIFF
--- a/deeprank2/neuralnets/cnn/model3d.py
+++ b/deeprank2/neuralnets/cnn/model3d.py
@@ -1,7 +1,7 @@
 import torch
-import torch.nn
-import torch.nn.functional as F
+from torch import nn
 from torch.autograd import Variable
+from torch.nn.functional import relu
 
 # ruff: noqa: ANN001, ANN201, ANN202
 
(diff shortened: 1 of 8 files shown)


gcroci2 (approved):
For being sure that Sphinx is able to solve the reference, you should use the fully qualified name (:class:`torch.nn.Module`). But it may be able to solve it also with the short version - not entirely clear. You should see after #625 is merged (now the trainer module is not correctly rendered) by activating the branch on readthedocs. Maybe just revert it for simplicity.

DaniBodor:
> For being sure that Sphinx is able to solve the reference, you should use the fully qualified name (:class:`torch.nn.Module`). But it may be able to solve it also with the short version - not entirely clear. You should see after #625 is merged (now the trainer module is not correctly rendered) by activating the branch on readthedocs. Maybe just revert it for simplicity.

I reverted it.

PR #646, opened 2024-09-16
docs: update CHANGELOG.md

gcroci2:
The CHANGELOG.md file should be kept updated, and the easiest way to do it is to update it manually every time a new release is made. Since the notes about the changes are automatically generated when we do a new release, I find that updating manually the file is unnecessary and redundant. If we like this, I will update accordingly https://github.com/ANNUBS/neurogym/pull/56 as well.


DaniBodor (approved):
Good idea, reduces redundancy.
I think this can be merged directly into main if you want

